Issue #3, opened 2021-04-08
Symptome_compartment_basic_SEIR_model


DSeiferth:
SEIR basic model for compartments of symptoms

PR #6, opened 2021-04-09
Add symptome-compartment model and test (#3)

hfldai:
Logs on branch [`i3-symptome-compartment`](https://github.com/Como-DTC-Collaboration/como-models/tree/i3-symptome-compartment)

### Model specificities
- [x] Simplify the infectious groups
- [x] Include unittest
- [x] Add disease-related deaths  
- [x] User-friendly naming for variables and parameters
- [x] Refine symptom compartments: asymptomatic, mild, severe 
- [x] Different lambdas for each of the mixings
- [x] Include R0 calculation for the system 
- approach: [next generation matrix](https://web.stanford.edu/~jhj1/teachingdocs/Jones-on-R0.pdf)
- [x]  Add `Incidence` and daily deaths into the output 
- [ ]  Include birth
- [ ]  Include natural deaths
- [ ] Thought: Rt calculation throughout the simulation period?

### Format (goal - unify model structure across classes; generalize common functions / class definitions):

- [x] Define a general `plot_dataframe` function outside of the model class (can be applied to any long-format dataframe)
- [x] Output dataframe in long format (as we did in plot function)
- [x] Restructure the model (rename slots, instances and params)
- [x] Moving `R0` and `output` incidences out from the class. The corresponding methods `R0()` and `run()` simply return values rather than updating the class.
- [ ] Thought: Move class declarations and related function definitions to objects.R
- [ ] Thought: Move all generic into a single .R file
- [ ] Thought: class inheritance

### Others
- [ ] RMD check fails when including `plotly` and `kableExtra` in package and vigenettes. -- WHY?
Parameter namings - reference list


DSeiferth (approved):
I ran "devtools::test()" and passed all tests (test_symptome_compartment_class.R)

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
+#' An SEIR model for compartments of different symptoms
+#'
+#' Note:
+#' 1) Sum of poulation equals to 1 at any time point.
```
ben18785: "population" -- I'd actually change this to be "Total population size is normalised to 1."

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#' An error will be reported the sum of initial population is not equal to 1.
```
ben18785: I'm not sure this needs to be here as it'll be handled in the code. Also, just about maintainability: if we removed that error check, we'd also have to remove this.

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#' 2) At this stage, we only consider disease-caused death cases
```
ben18785: I'm not sure I understand this comment? Do you mean that, at the moment, there are no natural deaths (only those from infected individuals)?

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#' calculating mortality.
+#'
+#' model structure:
+#'    ---------------------------(omega)---------------------------------
+#'    |                                                                 |
+#'    |                                                                 |
+#'    --> S --(lam)--> E --(gamma) ------(1-ihr)-------> I ---(nu.i)--> R <--
```
ben18785: I'm not sure I understand notation here. What's omega? What does CM or HA mean? I suspect it may be easier to represent things in equation form...

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#'                                 |            |                           |
+#'                                 |      (pdeath.hsa)                      |
+#'                                 |            |                           |
+#'                                 -- (ihr)---> HA --(nu.hsa)----------------
+
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+# Class definitions
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+#' @describeIn model_symptome to set up dependent classes for model_symptome
+setClass("ggplot")
```
ben18785: What do these two setClasses do?

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+setClass("gg")
+
+#' An S4 class of an basic SEIR model with compartments of different symptomes
```
ben18785: "symptoms" -- I think you can be a bit more specific here about classes with different symptoms. Is this the place to write down the model?

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#'
+#' @slot name A string gives the name of the model
+#' @slot initial_population A list of the initial population size of each group
```
ben18785: Is it a named list? I would have thought that would be a good idea to avoid writing to the wrong compartment.

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+# Class definitions
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+#' @describeIn model_symptome to set up dependent classes for model_symptome
+setClass("ggplot")
+setClass("gg")
+
+#' An S4 class of an basic SEIR model with compartments of different symptomes
+#'
+#' @slot name A string gives the name of the model
```
ben18785: I'm a bit confused. Why do we need to give this object a character name when we instantiate it?

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#' @param lam Numeric, rate at which an infected individual exposes susceptible
+#' @param gamma Numeric, rate at which exposed individuals become infected
+#' @param omega Numeric, rate at which recovered individuals become susceptible
+#' @param ihr Numeric, fraction of infected individuals develop severe symptoms
+#' @param nu Numeric vector, rate at which each infected group recover
+#' @param pdeath Numeric vector, rate of disease-caused mortality of each infected group
+#' @return An object of class model_symptome with initial population and parameters
+#'
+setGeneric(name = "set_init",
+           def = function(object, S=NA_real_, E=NA_real_, I=NA_real_,
```
ben18785: I would separate out the transmission parameters (lam, gamma etc.) from the initial compartment values (S_0, E_0 etc.) since they're two different types of object.

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+                  times = times,
+                  func = ode_symptome_rhs,
+                  parms = params,
+                  method = method)
+    output <- as.data.frame(output)
+    object@output <- output
+    return(object)
+    })
+
+#' @describeIn model_symptome to plot the outcome of the ode similuation
```
ben18785: I think this would probably work better as a separate plotting function as it could probably be written to take _any_ model type as input.

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+#' @slot output A dataframe holding the ode simulation output, with columns as population groups and rows as simulation time points
+#' @slot plot_output A ggplot holding the plot of the ode simulation output
+#'
+#' @import deSolve
+#' @import ggplot2
+#' @import reshape2
+#' @import plyr
+#' @import magrittr
+
+setClass(Class = "model_symptome",
```
ben18785: I'm not sure about this name for the class: I'd go for something more indicative like SEIaIsR (if that's the type of model it is, that is?).

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+    # ODE system RHS
+    ode_symptome_rhs <- function(t, pop_groups, parameters) {
+      with(
+        as.list(c(pop_groups, params)), {
+          dS <- -lam * S * (I + HA) + omega * R
+          dE <- lam * S * (I + HA) - gamma * E
+          dI <- (1.0 - ihr) * gamma * E - nu.i * I  - pdeath.i * I
+          dHA <- ihr * gamma * E - nu.hsa * HA  - pdeath.hsa * HA
+          dR <- -omega * R + nu.i * I + nu.hsa * HA
+          dCM <-  pdeath.i * I + pdeath.hsa * HA
```
ben18785: I think "I" is mild infection? I would have thought it might be more reflective of the Como model to have three compartments: `I_asymptomatic`, `I_mild`, `I_hospitalised` (I'd call them those names too so that people know what they are). Then you could just define `I_total = I_asymptomatic + I_mild + I_hospitalised` which you'd use in the `S` and `E` equations? Similarly, I'd call `CM` -> `D_cumulative` to make it more understandable.

ben18785 (changes requested):
Thanks @hfldai @DSeiferth -- looks like a great start. I do have quite a few thoughts about it though that I think would be good to discuss...

R/symptome_compartment_class.R
hfldai: Yes that's true - the comment refers to we do not consider natural deaths right now. Sorry for the confusion.

R/symptome_compartment_class.R
hfldai: In terms of naming of compartments, I totally agree that what you suggest are more readable and user-friendly.  Right now I am taking the names from the como code for both the compartments and the transmission parameters. So just to confirm, we don't necessarily want to keep the names of the compartments same with the current como code, do we?

R/symptome_compartment_class.R
hfldai: The S4 class model object contains a gg class object for storing a ggplot of the simulation outcome: 
`plot_output = "gg"`
The classes ggplot and gg need to be explicitly set as they don't belong to default object types.

R/symptome_compartment_class.R
hfldai: It is just to make a simple flowchart-like diagram for myself to check which population groups and parameters are included. Docstrings for the params and groups are further down in the code.  I meant to remove the part when the class is finished, but I agree it's better to represent by equations in docstrings. Will do it soon.

solveigvandervegt: I agree with Ben. I think this might be a great graph to have in the vignette that will go with this class though!

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+           parameters = vector(mode = "list", length = 6) %>%
+             setNames(list("lam", "gamma", "omega", "ihr", "nu", "pdeath")),
+           output = data.frame(),
+           plot_output = ggplot()
+         ))
+              
+#' @describeIn model_symptome set initial population sizes (in fraction) and parameters
+#'
+#' @param object An object of class model_symptome
+#' @param S Numeric, initial population of the susceptible
```
solveigvandervegt: maybe "initial fraction of the population that is susceptible" or something along those lines is a bit more consistent with a normalized population. If not, I think "initial population of susceptible" might be slightly better English.

R/symptome_compartment_class.R
```diff
@@ -0,0 +1,226 @@
...
+setGeneric(name = "check_init", def = function(object) standardGeneric("check_init"))
+setMethod("check_init", signature(object = "model_symptome"),
+          function(object) {
+            errors <- character()
+            # check whether all required parameters are set
+            is_na_params <- is.na(object@parameters)
+            if (sum(is_na_params) != 0) {
+              msg <- paste("Missing parameters:",
+                           paste(names(object@parameters)[is_na_params], collapse = ", ")
+              )
```
solveigvandervegt: I think we are just using the stop() function at the moment to throw errors. Is there a good reason to use one over the other?

solveigvandervegt (commented):
Looks good overall! Just some small extra comments from me, but mostly Ben has said everything

R/symptome_compartment_class.R
ben18785: Ok makes sense. I guess if we change the plotting to happen outside of the simulation function, this won't be the case though?

vignettes/SEIR_Model_Compartments.Rmd
```diff
@@ -0,0 +1,137 @@
...
+model <- set_init(object = model, S = S, E = E, I_asymptomatic = I_asymptomatic, I_mild = I_mild,
+                    I_severe = I_severe, R = R, D_cumulative = D_cumulative,
+                    lam = lam, gamma = gamma, omega = omega,
+                    e2i = e2i, i2r = i2r, pdeath = pdeath)
+```
+
+We use the ode solver and then plot the relaxation into the steady state:
+```{r}
+model <- ode_simulate(model, t)
+model <- plot_ode_output(model)
```
ben18785: I'm not sure why `model` is being assigned to the outcome of `plot_ode_output`?

vignettes/SEIR_Model_Compartments.Rmd
```diff
@@ -0,0 +1,137 @@
...
+
+## The SEIaImIsR Model
+The SEIaImIsR model consists of six ODEs describing the six populations. Susceptible
+individuals ($S$) have not been exposed to the virus. Exposed individuals ($E$) have been
+exposed to the virus and have been infected but are not yet infectious to others; they are 
+in the incubation period. Infectious individuals can spread the virus to others. 
+The total initial population size is normalised to 1.
+The current model does not include natural death or birth.	
+The ODEs describing how these six populations evolve over time are
+$$\frac{\text{d}S}{\text{d}t} = -\lambda S (I_{\text{a}} + I_{\text{m}} + I_{\text{s}}) + \omega R,$$
```
ben18785: So, at the moment, we assume that the mixing of the asymptomatics with the susceptibles is the same as say the severely infected? I think we probably want to allow different lambdas for each of the mixings as asymptomatics will probably continue mixing as normal; mild infected will probably isolate; and severely infected will be hospitalised.

vignettes/SEIR_Model_Compartments.Rmd
```diff
@@ -0,0 +1,137 @@
...
+where $\lamda$ denotes the rate, at which an infected individual exposes susceptible individuals,
+and $\omega$ denotes the rate, at which recovered individuals become susceptible.
+$$\frac{\text{d}E}{\text{d}t} = \lambda S (I_{\text{a}} + I_{\text{m}} + I_{\text{s}}) -\gamma E,$$
+where $\gamma$ denotes the rate at which exposed individuals become infected.
+$$\frac{\text{d}I_{\text{a}}}{\text{d}t} = (1-\alpha_{E->I_m} -\alpha_{E->I_s}) \gamma E - \beta_{I_a->R}  I_{\text{a}} - \mu_{I_a->D} * I_{a},$$
```
ben18785: I'd be tempted to have an alpha_{E->I_a} since there's always talk of the proportion of individuals who are asymptomatic; there's also talk of the hospitalisation rate of patients (i.e. those becoming severely infected), so I'd have a alpha_{E->I_s}. So, I'd be tempted to drop alpha_{E->I_m}.

vignettes/SEIR_Model_Compartments.Rmd
```diff
@@ -0,0 +1,137 @@
...
+individuals die due to the infection after $1/ \mu_i$ days, making the death rate $\mu_i$.
+
+In addition to these six main ODEs, the model also keeps track the number of 
+disease-related deaths ($D$) over time
+$$\frac{\text{d}D}{\text{d}t} = - \mu_{I_a->D} * I_{a} +  \mu_{I_m->D} * I_{m} +  \mu_{I_s->D} * I_{s},$$
+
+The system is closed by specifying the initial conditions
+$$S(0) = S_0,\ E(0) = E_0,\ I(0) = I_0,\ R(0) = R_0,\ C(0) = C_0,\ D(0) = D_0,$$
+In the implementation in this package, the variables are normalized so 
+$S(t) + E(t) + I_a(t) + I_m(t) + I_s(t) + R(t) + D(t) \equiv 1$ for any given $t$.
```
ben18785: I think this is right but have you checked this in the equations?

vignettes/SEIR_Model_Compartments.Rmd
```diff
@@ -0,0 +1,137 @@
...
+                    lam = lam, gamma = gamma, omega = omega,
+                    e2i = e2i, i2r = i2r, pdeath = pdeath)
+```
+
+We use the ode solver and then plot the relaxation into the steady state:
+```{r}
+model <- ode_simulate(model, t)
+model <- plot_ode_output(model)
+```
+
```
ben18785: Can we show how transmission changes as we change one / more of the parameters? I'd say, once we have it, allowing  different values for lambda_a, lambda_m and lambda_s would be interesting to compare to the lambda_a=lambda_m=lambda_m case, perhaps?

ben18785 (changes requested):
Just a few additional comments on the vignette... looks really good though!

vignettes/SEIR_Model_Compartments.Rmd
hfldai: In our design, all the functions taking the class SEIaImIsR as an input return also an S4 object in class SEIaImIsR, with modifications to the corresponding slots of the input . Here, `model` is an SEIaImIsR object with initial conditions and params set up.  `ode_simulate` runs the ode simulation and stores the output into the slot `model@output`, and `plot_ode_simulate` generates a ggplot for visualizing the output and stores the plot into `model@plot_output` .

vignettes/SEIR_Model_Compartments.Rmd
DSeiferth: Shall we add another model in the vignette, plot and compare it?

vignettes/SEIR_Model_Compartments.Rmd
hfldai: update: remove the `plot_ode_output` slot from the class, as the function has been moved outside of the class methods.

vignettes/SEIR_Model_Compartments.Rmd
hfldai: update: allowed different effective contact rates for each symptom compartment.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,294 @@
...
+#' are subgrouped into compartments according to different severity of symptoms, i.e. asymptomatic, mild and severe.
+#'
+#' 1. Total initial population size is normalised to 1.
+#' 2. The current model does not include natural death or birth.
+#' 3. The current model defines different infection.
+#'
+#' @slot initial_condition_names list of names of initial conditions
+#'       (characters). Default is list("S0", "E0", "I_asymptomatic0", "I_mild0", "I_severe0", "R0", "D0").
+#' @slot transmission_parameter_names list of names of transmission parameters
+#'       (characters). Default is list("beta", "kappa", "kappa", "mu").
```
FarmHJ: Two kappa here. And it doesn't match the default one listed in the class (line 52-52)

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,294 @@
...
+           R0 = "numeric",
+           output_names = "list",
+           output = "list"
+         ),
+         prototype = list(
+           initial_condition_names = list("S0", "E0", "I_asymptomatic0",
+                                          "I_mild0", "I_severe0", "R0", "D0"),
+           transmission_parameter_names = list("beta", "kappa", "omega",
+                                              "p_symptom", "gamma", "mu"),
+           initial_conditions = vector(mode = "list", length = 6),
```
FarmHJ: Minor comment, I think the length of initial_conditions should be 7.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,294 @@
...
+              omega = object@transmission_parameters$omega,
+              p_symptom = object@transmission_parameters$p_symptom,
+              gamma = object@transmission_parameters$gamma,
+              mu = object@transmission_parameters$mu)
+
+  # ODE system RHS
+  ode_symptome_rhs <- function(t, pop_groups, parameters) {
+    with(
+      as.list(c(pop_groups, params)), {
+        dS <- -S * (beta.i_asymptomatic * I_asymptomatic + beta.i_mild * I_mild + beta.i_severe * I_severe) + omega * R
```
FarmHJ: Are we including waning immunity in this model? I might have missed the discussion.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,294 @@
...
+#'
+set_init <- function(
+  object,
+  S=NA_real_,
+  E=NA_real_,
+  I_asymptomatic=NA_real_, I_mild=NA_real_, I_severe=NA_real_,
+  R=NA_real_,
+  D=NA_real_,
+  beta=list(), kappa=NA_real_, omega=NA_real_,
+  p_symptom=list(), gamma=list(), mu=list()
```
FarmHJ: It's not obvious how the named list of `p_symptom, gamma` and `mu` should be constructed for the function input. It might be better to show in some way what names to use for the lists.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,294 @@
...
+
+#' Solves the ode system.
+#'
+#' \deqn{\frac{dS(t)}{dt} = -S(t)(\beta.i_{asymptomatic} I_{asymptomatic}(t) + \beta.i_{mild} I_{mild}(t) + \beta.i_{severe} I_{severe}(t)) + \omega R(t)}
+#' \deqn{\frac{dE(t)}{dt} = S(t)(\beta.i_{asymptomatic} I_{asymptomatic}(t) + \beta.i_{mild} I_{mild}(t) + \beta.i_{severe} I_{severe}(t)) - \kappa E(t)}
+#' \deqn{\frac{dI_{asymptomatic}(t)}{dt} = p_symptom.i_{asymptomatic} \kappa E(t) - (\gamma.i_{asymptomatic} + \mu.i_{asymptomatic}) I_{asymptomatic}(t)}
+#' \deqn{\frac{dI_{mild}(t)}{dt} = p_symptom.i_{mild}\kappa E(t) - (\gamma.i_{mild} + \mu.i_{mild}) I_{mild}(t)}
+#' \deqn{\frac{dI_{severe}(t)}{dt} = p_symptom.i_{severe}\kappa E(t) - (\gamma.i_{severe} + \mu.i_{severe}) I_{severe}(t)}
+#' \deqn{\frac{dR(t)}{dt} = -\omega R(t) + \gamma.i_{mild} I_{mild}(t) + \gamma.i_{severe} I_{severe}(t)}
+#' \deqn{\frac{dD(t)}{dt} =  \mu.i_{mild} I_{mild}(t) + \mu.i_{severe} I_{severe}(t)}
```
FarmHJ: One term is missing here

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,199 @@
...
+we solve the ODEs over time instead of just solving for the steady states. The
+next section shows an example of how to use the SEIR_model class.
+
+### Basic reproduction number
+The mathematical details refer to [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6002118/pdf/main.pdf).
+
+We calculate the basic reproduction number ($\mathscr{R_0}$) using the next generation matrix approach:
+
+First, define 
+$x\in (x_1, x_2,...,x_n)^T$ be the number of individuals in each compartment, where the first $m < n$ compartments contain exposed or infected individuals. In the **SEIaImIsRD** model, these compartments $E$, $I_a$, $Im$ and $Is$).
```
FarmHJ: Missing underscore for variable Im and Is

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,199 @@
...
+where $\eta_{E->I_a}$, $\eta_{E->I_m}$, $\eta_{E->I_s}$ denote the fraction of exposed individuals that develop asymptomatic, mild and symptoms, respectively (in the code the fraction is denoted by 'p_symptom').
+The rate, at which each infected group recovers, is denoted by $\gamma_I$.
+The rate of disease-caused mortality of each infected group is denoted by $\mu_I$.
+
+$$\frac{\text{d}R}{\text{d}t} = - \omega R +\mu_{I_a}  * I_{a} + \mu_{I_m}  * I_{m} + \mu_{I_s}  * I_{s}.$$
+
+The rate at which susceptible individuals are infected depends on the fraction
+of the population that is susceptible, the fraction of the population that is
+infectious, and the probability of an infectious individual infecting a 
+susceptible individual, $\lambda$. This is represented in the
```
FarmHJ: It should be beta, not lambda

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,199 @@
...
+Exposed individuals become infectious at rate $\kappa$. The reciprocal value $1/ \kappa$ is
+the incubation time of the virus, i.e. how many days after exposure the individual
+becomes infectious to others. Infectious individuals can either recover or die
+because of the disease. Infectious individuals in group $i\in {a,m,s}$ recover in $1/ \beta_i$
+days, thus individuals recover at rate $\beta_i$. Similarly, infectious
+individuals die due to the infection after $1/ \mu_i$ days, making the death rate $\mu_i$.
+
+In addition to these six main ODEs, the model also keeps track the number of 
+disease-related deaths ($D$) over time
+$$\frac{\text{d}D}{\text{d}t} = - \mu_{I_a} * I_{a} +  \mu_{I_m} * I_{m} +  \mu_{I_s} * I_{s},$$
```
FarmHJ: You have an extra negative sign in the equation

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,199 @@
...
+$$S(0) = S_0,\ E(0) = E_0,\ I(0) = I_0,\ R(0) = R_0,\ C(0) = C_0,\ D(0) = D_0,$$
+In the implementation in this package, the variables are normalized so 
+$S(t) + E(t) + I_a(t) + I_m(t) + I_s(t) + R(t) + D(t) \equiv 1$ for any given $t$.
+
+Steady states of the system described above are computed by setting the right-hand
+sides of the ODEs equal to zero. 
+
+Because we build on this basic model to develop more complex epidemiological models,
+we solve the ODEs over time instead of just solving for the steady states. The
+next section shows an example of how to use the SEIR_model class.
```
FarmHJ: Typo, it's not SEIR_model class.

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,199 @@
...
+                    p_symptom = p_symptom, gamma = gamma, mu = mu)
+
+model1 <- ode_simulate(model1, t)
+model1 <- R0_SEIaImIsRD(model1)
+p1 <- plot_dataframe(model1@output, x = "time", y = "fraction", c = "population_group")
+print(p1)
+print(paste0("model1: R0=", model1@R0))
+```
+
+Comparison with altered beta values:
```
FarmHJ: Maybe add a short description on what beta is, like what you did for omega above.

FarmHJ (commented):
Great job! I have some comments, most of them are just typos. 
Do we want to calculate number of cases and deaths in each model, or include it into the generic plotting function? 
It's interesting to me that waning immunity is not considered when calculating the basic reproduction number. The course of the epidemic is so different with and without waning immunity, yet their reproduction numbers are the same.

R/SEIaImIsRD.R
hfldai: Yes, the `omega * R` term is introducing the loss of immunity in the recovered population

hfldai:
> Do we want to calculate number of cases and deaths in each model, or include it into the generic plotting function?

I think currently we calculate cumulative disease-caused deaths in each model separately, as each model consists of different infectious groups. The plotting function is defined outside of the model class, which can be called by different models, given they have the same format of output dataframe.

> It's interesting to me that waning immunity is not considered when calculating the basic reproduction number. The course of the epidemic is so different with and without waning immunity, yet their reproduction numbers are the same.

Yes, whether or not including waning immunity gives different epidemic curves even though at the same R0, as R0 is the basic reproduction number that only implies the trend at the beginning of the epidemic, under the disease free equilibrium (DFE) condition (S0>0, E0>0, S0>>E0, other compartments = 0); thus R0 does not imply how the pandemic behaves in the later times. I think If we monitor the effective reproduction number R(t) for each time point, we might see the differences caused by waning immunity (and maybe also the effects by other conditions/parameters).

R/symptome_compartment_class.R
ben18785: I think this is ok because these are now collected into a stop later on.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,356 @@
...
+
+#' @describeIn SEIaImIsRD Setter method for transmission parameters
+#' of the SEIaImIsRD model.
+#'
+#' @param object An object of class SEIaImIsRD
+#' @param value A numeric named list of values for transmission parameters:
+#' beta - a list of the effective contact rate from each infected group (i.e. rate at which an infected individual exposes susceptible),
+#' kappa - rate of progression from exposed to infectious (the reciprocal is the incubation period),
+#' omega - rate at which recovered individuals become susceptible,
+#' p_symptom - a list of fraction of different infected groups,
```
ben18785: What's the order of this list? Is it worth making this a named list?

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,356 @@
...
+    else stop(paste(errors, ", please check and rerun transmission_parameters<-.\n"))
+  })
+
+
+#' @describeIn SEIaImIsRD Setter method for transmission parameters
+#' of the SEIaImIsRD model.
+#'
+#' @param object An object of class SEIaImIsRD
+#' @param value A numeric named list of values for transmission parameters:
+#' beta - a list of the effective contact rate from each infected group (i.e. rate at which an infected individual exposes susceptible),
```
ben18785: Is it worth making this a named list?

R/SEIaImIsRD.R
ben18785: Also, does it sum to 1? It may be worth stating this explicitly.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,356 @@
...
+    })
+
+#' Solves the ode system.
+#'
+#' \deqn{\frac{dS(t)}{dt} = -S(t)(\beta.i_{asymptomatic} I_{asymptomatic}(t) + \beta.i_{mild} I_{mild}(t) + \beta.i_{severe} I_{severe}(t)) + \omega R(t)}
+#' \deqn{\frac{dE(t)}{dt} = S(t)(\beta.i_{asymptomatic} I_{asymptomatic}(t) + \beta.i_{mild} I_{mild}(t) + \beta.i_{severe} I_{severe}(t)) - \kappa E(t)}
+#' \deqn{\frac{dI_{asymptomatic}(t)}{dt} = p_symptom.i_{asymptomatic} \kappa E(t) - (\gamma.i_{asymptomatic} + \mu.i_{asymptomatic}) I_{asymptomatic}(t)}
+#' \deqn{\frac{dI_{mild}(t)}{dt} = p_symptom.i_{mild}\kappa E(t) - (\gamma.i_{mild} + \mu.i_{mild}) I_{mild}(t)}
+#' \deqn{\frac{dI_{severe}(t)}{dt} = p_symptom.i_{severe}\kappa E(t) - (\gamma.i_{severe} + \mu.i_{severe}) I_{severe}(t)}
+#' \deqn{\frac{dR(t)}{dt} = -\omega R(t) + \gamma.i_{mild} I_{mild}(t) + \gamma.i_{severe} I_{severe}(t)}
```
ben18785: Shouldn't asymptomatics also recover?

ben18785: Think this is just a typo as it's in the ODE itself?

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,356 @@
...
+#' @rdname plot_dataframe
+#' @export
+#' @aliases plot_dataframe,ANY,ANY-method
+#'
+plot_dataframe <- function(dataframe, x = "time", y = "value", c = "compartment") {
+  p <- ggplot(dataframe, aes_string(x = x, y = y)) +
+    geom_line(aes_string(colour = c)) +
+    theme_classic()
+  return(p)
+}
```
ben18785: I'm not sure we need this as it's basically just wrapping a standard ggplot call.

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,261 @@
...
+
+The number of mildly asymptomatic infectious individuals evolves over time according to
+$$\frac{dI_{m}}{dt} = \eta_{E->I_m} \kappa E - (\gamma_{I_m} + \mu_{I_a})I_{m},$$
+where $\eta_{E->I_M}$, $\gamma_{I_M}$ and $\mu_{I_M}$ are positive parameters. $\eta_{E->I_a}$ is the fraction of exposed individuals that, after the incubation period, became infectious with mild symptoms. $\gamma_{I_M}$ denotes the rate at which mildly symptomatic individuals recover from the disease and $\mu_{I_M}$ denotes the rate at which they die of the disease.
+
+The number of severely asymptomatic infectious individuals evolves over time according to
+$$\frac{dI_{s}}{dt} = \eta_{E->I_s} \kappa E - (\gamma_{I_s} + \mu_{I_a})I_{s},$$
+where $\eta_{E->I_S}$, $\gamma_{I_S}$ and $\mu_{I_S}$ are positive parameters. $\eta_{E->I_a}$ is the fraction of exposed individuals that, after the incubation period, became infectious with severe symptoms. $\gamma_{I_S}$ denotes the rate at which severely symptomatic individuals recover from the disease and $\mu_{I_S}$ denotes the rate at which they die of the disease.
```
ben18785: Can we perhaps condense these sections into a single one since the descriptions are quite repetitive?

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,261 @@
...
+
+The number of recovered individuals evolves over time according to
+$$\frac{\text{d}R}{\text{d}t} =  \mu_{I_a}  I_{a} + \mu_{I_m}  I_{m} + \mu_{I_s}  I_{s} - \omega R,$$
```
ben18785: Are the parameter names consistent between here and the documentation for the functions?

ben18785 (changes requested):
Thanks @hfldai -- looks really good. A few small things then should be ready to go!

R/SEIaImIsRD.R
hfldai: Yes it is a typo in the documentations. I've fixed it. the actual functions were correct.

R/SEIaImIsRD.R
hfldai: In p_symptom users only need to specify mild and severe groups. And the sum of the two should be <=1. I've included that in checks and docstrings.

R/SEIaImIsRD.R
hfldai: Moved to tests and vignettes.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,352 @@
...
+
+#' Solves the ode system.
+#'
+#' \deqn{\frac{dS(t)}{dt} = -S(t)(\beta.i_{asymptomatic} I_{asymptomatic}(t) + \beta.i_{mild} I_{mild}(t) + \beta.i_{severe} I_{severe}(t)) + \omega R(t)}
+#' \deqn{\frac{dE(t)}{dt} = S(t)(\beta.i_{asymptomatic} I_{asymptomatic}(t) + \beta.i_{mild} I_{mild}(t) + \beta.i_{severe} I_{severe}(t)) - \kappa E(t)}
+#' \deqn{\frac{dI_{asymptomatic}(t)}{dt} = p_symptom.i_{asymptomatic} \kappa E(t) - (\gamma.i_{asymptomatic} + \mu.i_{asymptomatic}) I_{asymptomatic}(t)}
+#' \deqn{\frac{dI_{mild}(t)}{dt} = p_symptom.i_{mild}\kappa E(t) - (\gamma.i_{mild} + \mu.i_{mild}) I_{mild}(t)}
+#' \deqn{\frac{dI_{severe}(t)}{dt} = p_symptom.i_{severe}\kappa E(t) - (\gamma.i_{severe} + \mu.i_{severe}) I_{severe}(t)}
+#' \deqn{\frac{dR(t)}{dt} = -\omega R(t) + \gamma.i_{asymptomatic} I_{asymptomatic}(t) + \gamma.i_{mild} I_{mild}(t) + \gamma.i_{severe} I_{severe}(t)}
+#' \deqn{\frac{dD(t)}{dt} =  \mu.i_{asymptomatic} I_{asymptomatic}(t) + \mu.i_{mild} I_{mild}(t) + \mu.i_{severe} I_{severe}(t)}
```
ben18785: I'm not sure we need the double subscripts. How about just (say) \beta_{mild} opposed to \beta.i_{mild}? Similarly for all the parameters that vary according to symptom group.

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+
+
+The SEIaImIsRD model consists of an ordinary differential equation (ODE) system of six ODEs, each describing the rate of the size changing of a population group evolves over time:
+
+
+$$\frac{\text{d}S}{\text{d}t} = -S (\beta_{I_a}I_a + \beta_{I_m}I_m + \beta_{I_s}I_s) + \omega R,$$
+$$\frac{\text{d}E}{\text{d}t} = S (\beta_{I_a}I_a + \beta_{I_m}I_m + \beta_{I_s}I_s) - \kappa E,$$
+$$\frac{dI_{a}}{dt} = \eta_{E->I_a} \kappa E - (\gamma_{I_a} + \mu_{I_a})I_{a},$$
+$$\frac{dI_{m}}{dt} = \eta_{E->I_m} \kappa E - (\gamma_{I_m} + \mu_{I_a})I_{m},$$
+$$\frac{dI_{s}}{dt} = \eta_{E->I_s} \kappa E - (\gamma_{I_s} + \mu_{I_a})I_{s},$$
```
ben18785: These parameter names don't currently match those in the documentation. Can we make these the same across both?

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+* $\beta_{I_i}$ - the rate at which an infected individual of symptom group $i$ infects susceptible individuals ($i\in\{a, m, s\}$).
+* $\omega$ - the rate at which recovered individuals become susceptible again.
+* $\kappa$ - the rate at which exposed individuals become infectious. The incubation period thus lasts $1/\kappa$ days.
+* $\eta_{E->I_i}$ - the fraction of exposed individuals that, after the incubation period, became infectious of symptom group $i$. $\sum_i\eta_{E->I_i}=1$, $i\in\{a, m, s\}$.
+* $\gamma_{I_i}$ - the rate at which individuals of symptom group $i$ recover from the disease ($i\in\{a, m, s\}$).
+* $\mu_{I_i}$ - the rate at which individuals of symptom group $i$ die of the disease ($i\in\{a, m, s\}$).
+
+In addition to these six main ODEs above, the model also keeps track the number of 
+disease-related deaths ($D$) and the total number of cases ($C$) over time: $$\frac{\text{d}D}{\text{d}t} = \mu_{I_a} I_{a} +  \mu_{I_m}I_{m} +  \mu_{I_s} I_{s}$$, 
+$$\frac{\text{d}C}{\text{d}t} = S (\beta_{I_a}I_a + \beta_{I_m}I_m + \beta_{I_s}I_s)$$.
```
ben18785: I think the full stop "." needs to be written within the dollar signs for the document to render properly.

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+with the following parameters (values between 0 and 1 for all):
+* $\beta_{I_i}$ - the rate at which an infected individual of symptom group $i$ infects susceptible individuals ($i\in\{a, m, s\}$).
+* $\omega$ - the rate at which recovered individuals become susceptible again.
+* $\kappa$ - the rate at which exposed individuals become infectious. The incubation period thus lasts $1/\kappa$ days.
+* $\eta_{E->I_i}$ - the fraction of exposed individuals that, after the incubation period, became infectious of symptom group $i$. $\sum_i\eta_{E->I_i}=1$, $i\in\{a, m, s\}$.
+* $\gamma_{I_i}$ - the rate at which individuals of symptom group $i$ recover from the disease ($i\in\{a, m, s\}$).
+* $\mu_{I_i}$ - the rate at which individuals of symptom group $i$ die of the disease ($i\in\{a, m, s\}$).
+
+In addition to these six main ODEs above, the model also keeps track the number of 
+disease-related deaths ($D$) and the total number of cases ($C$) over time: $$\frac{\text{d}D}{\text{d}t} = \mu_{I_a} I_{a} +  \mu_{I_m}I_{m} +  \mu_{I_s} I_{s}$$,
```
ben18785: Comma needs to be within the dollar signs.

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+
+We calculate the basic reproduction number ($R_0$) using the next generation matrix approach. First, we select those populations in the model that contain exposed or infectious individuals, so $E$, $I_a$, $I_m$ and $I_s$. We define $x = [E, I_a, I_m, I_s]$ and rewrite the ODEs of the variables in $x$ to be of the form $$\frac{\text{d}x_i}{\text{d}t} = \mathscr{F_i}(x) - \mathscr{V_i}(x)$$, where $\mathscr{F_i}(x)$ is the rate of appearance of new infections in compartment $i$, and $\mathscr{V_i}(x)$ is the rate of other transitions between compartment $i$ and other infected compartments. For example, if we take $x_i = E$, then $\mathscr{F_i}(x) = S (\beta_{I_a}I_a + \beta_{I_m}I_m + \beta_{I_s}I_s)$ and $\mathscr{V_i}(x) = \kappa E$. Similarly, for $x_i = I_a$ $\mathscr{F_i}(x) = \eta_{E->I_a} \kappa E$ and $\mathscr{V_i}(x) = (\gamma_{I_a} + \mu_{I_a})I_{a}$.
```
ben18785: For the dxi_dt equation, can we put blank between it and the text otherwise it's hard to see where it appears. Also, can we put the comma within the dollar signs

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+
+where the first three terms indicate recovery of infectious individuals with different levels of symptoms and the last term indicates loss of immunity.
```
ben18785: It's not clear where this line refers to? It also doesn't seem to be a correct sentence.

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,352 @@
...
+  "transmission_parameters<-",
+  "SEIaImIsRD",
+  function(
+    object,
+    value = list(beta = list(i_asymptomatic = NA_real_, i_mild = NA_real_, i_severe = NA_real_), 
+    kappa = NA_real_, 
+    omega = NA_real_,
+    p_symptom = list(i_mild = NA_real_, i_severe = NA_real_), 
+    gamma = list(i_asymptomatic = NA_real_, i_mild = NA_real_, i_severe = NA_real_), 
+    mu = list(i_asymptomatic = NA_real_, i_mild = NA_real_, i_severe = NA_real_))) {
```
ben18785: Similar to the comment about notation below, can we just make the setters: beta = list(asymptomatic=, mild=, severe=)? In other words, drop the "i_" in all cases.

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+# print_table_transmission_parameters <- function(model){
+#   df.params <- model@transmission_parameters %>%
+#     data.frame() %>%
+#     t()
+#   colnames(df.params) <- "value"
+#   table.params <- kable(df.params) %>%
+#     kable_styling(bootstrap_options = "striped", full_width = FALSE, position="left")
+#   return (table.params)
+# }
+```
```
ben18785: If this isn't used, can we remove it?

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,352 @@
...
+
+#' @describeIn SEIaImIsRD Setter method for transmission parameters
+#' of the SEIaImIsRD model.
+#'
+#' @param object An object of class SEIaImIsRD
+#' @param value A numeric named list of values for transmission parameters:
+#' beta - a named list of the effective contact rate from each infected group (i.e. rate at which an infected individual exposes susceptible), each element with value in [0,1]
+#' kappa - rate of progression from exposed to infectious (the reciprocal is the incubation period),
+#' omega - rate at which recovered individuals become susceptible,
+#' p_symptom - a named list of fraction of different infected groups, only need to specify values for mild and severe symptom groups, whose sum should be <=1.
```
ben18785: I think this could perhaps be explained a bit better: shouldn't it be the probability of exposed individuals moving into each of the different infected groups? Also, can we also say that, "Here, only probabilities for the mild (p_m) and severe (p_s) groups need be specified and the asymptomatic probability is the remainder (1 - p_m - p_s)."

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+# get parameters
+# model.params.table = print_table_transmission_parameters(model)
+
+# <div class = "row">
+# <div class = "col-md-3">
+# model.params.table
+# </div>
+# <div class = "col-md-8">
+# </div>
+# </div>
```
ben18785: This isn't used as far as I can tell, so can go.

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+Here, it makes sense to go with a time step of one as data for an epidemic, such as number of cases and deaths, is often reported per day, so time resolution of hours or minutes is not useful.
+
+The system of ODEs is solved by calling the function `run(model,t)`. In the next few sections, we run the model under different assumption and plot the results. The function `R0(model)` can be used to compute the basic reproduction number for the model with a given set of parameters. 
+
+#### 1. Including natural loss of immunity
+We first assume that immunity does not last forever, so individuals that have recovered from the disease eventually return to the susceptible population. This is called waning immunity and is included in the model when $\omega>0$, as we assigned it above.
+```{r, fig.height = 5, fig.width = 5, fig.align = "center", fig.show="hold"}
+par(mfrow=c(1,2))
+model <- run(model, t)
+model <- R0(model)
```
ben18785: It's confusing that you rename model to be the new R0? Also, perhaps put the call to R0 in the next block near where it's printed?

vignettes/SEIRD_Model_Compartments.Rmd
```diff
@@ -0,0 +1,249 @@
...
+We see that for this set of parameter values, $R_0>1$, and indeed we see in the plot that the epidemic initially grows. 
+
+#### 2. No natural loss of immunity 
+We now assume that recovered individuals remain immune forever, i.e. $\omega=0$.
+```{r,  fig.height = 5, fig.width = 5, fig.align = "center", fig.show="hold"}
+model1 <- model
+transmission_parameters(model1) <- list(beta = beta, kappa = kappa, omega = 0.00,
+                                        p_symptom = p_symptom, gamma = gamma, mu = mu)
+model1 <- run(model1, t)
+model1 <- R0(model1)
```
ben18785: Similarly here with calling model1 the R0

ben18785 (changes requested):
Hi @hfldai -- looks really good, thanks. Apologies but there are still a few things that need to be sorted before this goes in. One thing in addition to my comments is, can we rename the vignette file? Rather than "SEIRD_Model_Compartments", perhaps it should have the same name as the model itself?

vignettes/SEIRD_Model_Compartments.Rmd
hfldai: No, it is not renamed. `R0()` is a class-specific function that returns an SEIaImIsRD object. Then `model@R0` get the actual value of R0. As a formal class definition, I set the all the actual functions for the class to return an object of the class, except for validation functions.

vignettes/SEIRD_Model_Compartments.Rmd
hfldai: It's for simplification when writing a lot of them. I mentioned them here: 
```
  * **Asymptomatic infectious** ($I_{asymptomatic}$, or $I_a$ for short);
  * **Mildly symptomatic infectious** ($I_{mild}$, or $I_m$ for short),
  * **Severely symptomatic infectious** ($I_{severe}$, or $I_s$ for short).
```

R/SEIaImIsRD.R
```diff
@@ -0,0 +1,344 @@
...
+
+            # Split output into 2 dataframes: one with S,E,I, and R and one with C and D
+            states <- subset(output, !output$compartment %in% c("Incidence", "Deaths"))
+            states <- droplevels(states)
+            changes <- subset(output, output$compartment %in% c("Incidence", "Deaths"))
+            changes <- droplevels(changes)
+            # reshape data frame into long format
+            # output <- melt(output, id.vars = "time")
+            # names(output) <- c("time", "population_group", "fraction")
+            # names(output) <- c("time", "compartment", "value")
```
ben18785:
```suggestion
```

Issue #1, opened 2021-03-31
Write SEIR model


FarmHJ:
The SEIR model is the base model for the full CoMo model.

FarmHJ:
- [x] Change format of code to S4 Class
- [x] Add unit testing
- [ ] Include vignettes to show the use of the function

Should we remove the figure and put that in another function?

PR #15, opened 2021-04-23
I1 seir model

FarmHJ:
Addresses #1. Creates a class for SEIDR model. Unit tests, documentations and vignettes are included.


R/SEIRD.R
```diff
@@ -0,0 +1,347 @@
...
+      2:length(output$D)] - output$D[1:(length(output$D) - 1)]
+
+    colnames(output) <- c("time", object@output_names)
+
+    # Create long format of output
+    output <- melt(output, id.vars = "time")
+    names(output) <- c("time", "compartment", "value")
+
+    # Added for consistency of output format across models
+    output$age_group <- rep("0-150", length(output$time))
```
oscardm20994: We should ensure the long format of the output is consistent with the SEIRAge model, this looks good but we've done it slightly differently. both would probably work well.

oscardm20994 (commented):
This is looking really good, @solveigvandervegt and @FarmHJ. Just need to make sure there is consistency across this and the age model in outputs etc. Other than that looks in really good shape.

R/SEIRD.R
solveigvandervegt: Thanks Oscar, I will change ours to match yours!

solveigvandervegt:
@ben18785 I think I made all the changes that we talked about. Let me know if I missed anything!

ben18785 (approved):
Great work guys! I'm happy to merge now.

Issue #23, opened 2021-07-15
Correct typo in simple SEIRD documentation


ben18785:
At the moment, it says "SEIR" opposed to SEIRD in one part of the documentation.

PR #24, opened 2021-07-15
Series of changes to docstrings for SEIRD

ben18785:
## Issue
- [x] Which issue does your pull request address?

#23 

## Testing, documentation, and checks
- [x] If your pull request introduces new code, you should write automated tests (explained [here](https://r-pkgs.org/tests.html)).
- [x] Check that all tests pass using `devtools::check()`.
- [x] Build documentation using `devtools::document()`.

### DIFF
--- a/R/SEIRD.R
+++ b/R/SEIRD.R
@@ -1,10 +1,10 @@
-#' An S4 object representing the SEIRD.
+#' An S4 object representing the SEIRD model.
 #'
-#' This class represents the SEIR model, showing how populations of susceptible,
+#' This class represents the SEIRD model, which can be used to determine how populations of susceptible,
 #' exposed, infectious and recovered individuals evolve over time.
 #'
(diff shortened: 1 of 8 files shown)


solveigvandervegt:
@ben18785 I fixed some lint issues, but the rest looks good to me. One problem I see: you changed the description of the generic for run() to say that we return one data frame with S,E,I,R,D and time and one with incidences, deaths and time. You didn't change this in the description for the actual method (it still says one df with S,E,I and R and one with incidences and deaths) and looking at the code, I also don't think this is right (the description with the method is correct, I think). I might be wrong though, so have a look! It should at least be consistent between the generic and the method.

solveigvandervegt:
Actually, now that I think about it, if the generic is going to apply to all classes whereas the method is class specific, shouldn't the description of the generic be much more general so as to cover all cases?? Or will that mess things up?

ben18785:
This has been superceded by #50

Issue #21, opened 2021-07-01
Moving all Generics into generics.R after merging branches


hfldai:
As pointed in the answer to @I-Bouros 's [stackoverflow question](https://stackoverflow.com/questions/68113799/how-can-you-create-multiple-s4-classes-with-similarly-named-methods-in-r), duplicated declarations of generics are not allowed in R. To avoid ambiguity structurally, after merging the model branches we can 
1. move all generics declarations (`setGeneric.(..)`) into `generics.R`,
2. add `#' @include generics.R` to the front of each model R file, and `setMethod` can be used as usual.

An extra step might be documenting all shared generics into the same .Rd file in man (e.g. `generics.Rd`), while keeping class-specific generics in their class-specific Rd files.

solveigvandervegt:
We also need to change the documentation of the generics function so they apply to all models. Currently the documentation for the generics refers to the basic SEIRD model specifically.

ben18785:
Merged by #37

PR #25, opened 2021-07-27
I16 urban vs rural communities

solveigvandervegt:
pull request for the 2 communities model to be added to the como-models package.

Some issues remain around loading the excel files and I can therefor also not check if all the tests actually pass.


DESCRIPTION
```diff
@@ -19,6 +19,9 @@ Suggests:
 Config/testthat/edition: 3
 Imports: 
     deSolve,
+    matlib,
+    readxl,
+    gridExtra
```
rccreswell:
```suggestion
    gridExtra,
```

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+```{r setup}
+library(comomodels)
+library(ggplot2)
+library(glue)
+library(readxl)
+library(deSolve)
+library(reshape2)
+library(gridExtra)
+library(readxl)
+library(viridis)
```
rccreswell, line 26: Do we need to add viridis to Imports in DESCRIPTION?

DESCRIPTION
```diff
@@ -19,6 +19,9 @@ Suggests:
 Config/testthat/edition: 3
 Imports: 
     deSolve,
+    matlib,
+    readxl,
```
rccreswell: Since you converted the excel files, do we still need readxl?

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+```{r setup}
+library(comomodels)
+library(ggplot2)
+library(glue)
+library(readxl)
```
rccreswell:
```suggestion
```

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+library(deSolve)
+library(reshape2)
+library(gridExtra)
+library(readxl)
```
rccreswell:
```suggestion
```
Assuming we don't need these anymore

R/SEIRD_RU.R
```diff
@@ -0,0 +1,694 @@
...
+#'       (characters). Default is list("C_U0", "D_U0", "C_Y0", "D_Y0").
+#' @slot transmission_parameter_names list of names of transmission parameters
+#'       (characters). Default is list("b", "k", "g", "m", "C").
+#' @slot initial_conditions list of values for initial conditions (double).
+#' @slot initial_cases_deaths list of values for initial cases and deaths.
+#'       Both set to 0, not to be changed by user (double).
+#' @slot transmission_parameters list of values for transmission parameters
+#'       (double).
+#' @slot contact_matrices list of two contact matrices, one for urban communties
+#'       and one for rural communities, normalized to population size (double).
```
rccreswell: Not 100% sure what you mean by "normalized to population size" for the contact matrices

R/SEIRD_RU.R
```diff
@@ -0,0 +1,694 @@
...
+    # add names to each value
+    names(demo_data) <- object@country_demog_names
+
+    # check format of contact matrices
+    if (typeof(demo_data$urban) != "double"
+        | typeof(demo_data$rural) != "double") {
+      stop("Contact matrices must be of type double.")
+    }
+
+    # check that vectors are normalized: loosing some precision in division
```
rccreswell:
```suggestion
    # check that vectors are normalized: losing some precision in division
```

R/SEIRD_RU.R
```diff
@@ -0,0 +1,694 @@
...
+                (country_demog(object)$urban/sum(country_demog(object)$urban))) / 2
+    # number of rural contacts for a rural individual, normalized over the
+    # whole population through the age demographics
+    N_Y <- sum(rowSums(contact_matrices(object)$rural) *
+                (country_demog(object)$rural/sum(country_demog(object)$rural))) / 2
+
+    # fraction of the population that is urban
+    f_urban <- sum(country_demog(object)$urban)
+    #fraction of the population that is rural
+    f_rural <- sum(country_demog(object)$rural)
```
rccreswell: It's great that the class provides `country_demog<-` and `contact_matrices<-` and the computations here, as they make it very easy to work with real data. However, I would strongly favor adding an option so that users can alternatively provide the values of N_U, N_Y, phi_U, and phi_Y directly, in case they want to run simulations with particular values in mind. 
We can probably achieve this using the same methods, but either in the docstrings or the vignette we should tell the user exactly what inputs to send to those methods if they just want to provide 4 numbers N_U, N_Y, phi_U, phi_Y.

R/SEIRD_RU.R
```diff
@@ -0,0 +1,694 @@
...
+#'
+#' @return object of class SEIRD_RU with transmission parameter values
+#' assigned.
+#' @export
+
+setGeneric(
+  "transmission_parameters<-",
+  function(object, value) {
+    standardGeneric("transmission_parameters<-")
+  })
```
rccreswell: I think that these setGenerics will cause problems for the other classes. 
Based on https://github.com/Como-DTC-Collaboration/como-models/pull/5/commits/2cafce9eb38b9865f8144569b7e119b56eef2d37
https://github.com/Como-DTC-Collaboration/como-models/issues/21

Can you check that if you remove the setGeneric for transmission_parameters, run, and so forth, keeping only those for methods that are not included in any of the other classes, your model will still work?

R/SEIRD_RU.R
```diff
@@ -0,0 +1,694 @@
...
+    names(output) <- c("time", "compartment", "value")
+
+    # Added for consistency of output format across models
+    output$age_group <- rep("0-150", length(output$time))
+
+    return(output)
+  })
+
+#----------------------------------------------------------------------------
+#' Returns the value of R0 for the model with the specified parameter values.
```
rccreswell: Somewhere in the comments for this method, let's spell out that by R0 we mean "basic reproduction number"

R/SEIRD_RU.R
```diff
@@ -0,0 +1,694 @@
...
+      NY <- sum(rowSums(contact_matrices(object)$rural) *
+                   (country_demog(object)$rural/sum(country_demog(object)$rural))) / 2
+      # fraction of the population that is urban
+      f_urban <- sum(country_demog(object)$urban)
+      #fraction of the population that is rural
+      f_rural <- sum(country_demog(object)$rural)
+      # convenience parameter
+      K <- ((1-f_rural)* NU + f_rural*NY)
+      
+      matF <- matrix(data = c(0, 0, b*S0U*(K*C + NU/(1-f_rural)*(1-C)),              b*S0U*K*C,
```
rccreswell: The arithmetic is quite complicated here. While there's no need to repeat the entire method if we cite it, can we:
- Wherever is typical for R code, name the same paper you have in the vignette
- Include a brief comment saying what F and V mean, and that matF and matV are where you are putting the derivatives
- Indicate what ordering you have adopted of the state variables (I_U, E_U, ...) for your matrices

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+## Introduction
+Here we consider an extension to the basic SEIRD model that considers multiple interacting communities. For simplicity, we consider only two interacting communities here. Each community consists of a susceptible, exposed, infectious and recovered population. Although one can consider two generic communities, in this implementation we consider specifically one urban and one rural community, and use data from the WHO and UN to specify contacts in each community. In the following sections, we will explain the equations of the model, the implementation in this package and how to use it, and discuss briefly how the connectedness level between communities affects the results.
+
+## The mathematical model
+### The equations
+We base our model of connected communities on the model presented by Tun et al. [-@Tun21]. We designate the urban community $U$ and the rural community $Y$. Following the same naming pattern for variables and parameters as used for the basic SEIRD model, the ordinary differential equations describing how the populations evolve are as follows, where variable $X_Z$ designates the population of individuals with infectedness status $X$ in community $Z$,
```
rccreswell:
```suggestion
We base our model of connected communities on the model presented by Tun et al. [-@Tun21]. We designate the urban community $U$ and the rural community $Y$. Following the same naming pattern for variables and parameters as used for the basic SEIRD model, the ordinary differential equations describing how the populations evolve are as follows, where variable $X_Z$  (for $X \in \{S,E,I,R\}, Z \in \{U, Y\}$) designates the population of individuals with infectedness status $X$ in community $Z$,
```
Something like this could be nice so the new variables don't confuse anyone

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+    \frac{\text{d}I_U}{\text{d}t} = & \kappa E_U - (\gamma + \mu)I_U ,\\
+    \frac{\text{d}I_Y}{\text{d}t} = & \kappa E_Y - (\gamma + \mu)I_Y,\\
+    \frac{\text{d}R_U}{\text{d}t} = & \gamma I_U,\\
+    \frac{\text{d}R_Y}{\text{d}t} = & \gamma I_Y,\\
+    \frac{\text{d}D_U}{\text{d}t} = & \mu I_U,\\
+    \frac{\text{d}D_Y}{\text{d}t} = & \mu I_Y.\\
+\end{split}
+\end{equation}
+
+\noindent In the above equations, $\phi_U$ is the fraction of the population that lives in an urban environment, $\phi_Y$ is the fraction of the population that lives in a rural environment, $N_U$ is the average number of contacts that an urban individual has in a day (normalized over the whole population), and $N_Y$ is the average number of contacts that a rural individual has in a day (normalized over the whole population), so $(\phi_U N_U + \phi_Y N_Y)$ is the normalized average number of contacts for any individual. $C$ is the connectedness parameter, where $C=1$ means the communities are well-mixed with each other, and $C=0$ means the two communities do not interact at all.
```
rccreswell: Probably related to another comment---"normalize" is a bit vague, if you could make these definitions slightly more specific (maybe including an equation if necessary), that would be helpful

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
...ther community. The extent to which the later happens is determined by the connectedness parameter $C$. If the communities are completely separate, i.e. $C=0$, the probability that a susceptible individual in community $Z$ interacts with an infectious individual is the number of contacts that someone in that community has, $N_Z$, multiplied by the probability that a contact is infectious, $I_Z/\phi_Z$. If the communities are well-mixed with each other, the probability is again the number of contacts multiplied by the probability that a contact is infectious, although now the number of contacts that an individual has is averaged over the communities, so $\phi_U N_U + \phi_Y N_Y$, and the individual can encounter people from either community so the probability of an infectious contact is $I_U+I_Y$. The extent to which the communities are connected, designated by $C$, determines to which extent the two extreme contact patterns play a role, and thus we obtain the weighted average that appears in the equations above which designates the total probability of a susceptible individual in $Z$ having an infectious contact: $(I_U+I_Y) (\phi_U N_U + \phi_Y N_Y) C + \frac{I_Z}{\phi_Z}N_Z(1-C)$.
```
rccreswell:
```suggestion
The difference between this model and the basic SEIRD model is the computation of the number of infectious contacts that a susceptible individual has. This is reflected in the equations for $S_Z$ and $E_Z$. Individuals in either community can either come into contact with people from their own community or from the other community. The extent to which the latter happens is determined by the connectedness parameter $C$. If the communities are completely separate, i.e. $C=0$, the probability that a susceptible individual in community $Z$ interacts with an infectious individual is the number of contacts that someone in that community has, $N_Z$, multiplied by the probability that a contact is infectious, $I_Z/\phi_Z$. If the communities are well-mixed with each other, the probability is again the number of contacts multiplied by the probability that a contact is infectious, although now the number of contacts that an individual has is averaged over the communities, so $\phi_U N_U + \phi_Y N_Y$, and the individual can encounter people from either community so the probability of an infectious contact is $I_U+I_Y$. The extent to which the communities are connected, designated by $C$, determines to which extent the two extreme contact patterns play a role, and thus we obtain the weighted average that appears in the equations above which designates the total probability of a susceptible individual in $Z$ having an infectious contact: $(I_U+I_Y) (\phi_U N_U + \phi_Y N_Y) C + \frac{I_Z}{\phi_Z}N_Z(1-C)$.
```

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+First, we load the contact matrices for the urban and rural communities. Because the two sets of contact matrices don't cover the exact same countries, we also make a list of countries for which we have both contact matrices, which we can use later to check if the choice of country is valid.
+```{r}
+# load contact matrices
+load("../Data/contact_all_urban.rdata") # ROWS are age of person, COLUMNS are age of contact
+names_urban <- names(contact_all_urban)
+load("../Data/contact_all_rural.rdata")
+names_rural <- names(contact_all_rural)
+names_common <- intersect(names_urban,names_rural)
+```
+We also load in all demographic data, and load in a list showing the which three-letter code for each country.
```
rccreswell:
```suggestion
We also load in all demographic data, and load in a list showing the three-letter code for each country.
```

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+```
+
+Now that we have specified a country, we extract from the data what percentage of its population is rural. Because the most recent year that all our data source have in common is 2019, we select all our data from that year.
+```{r}
+# get fraction of population that is rural
+country_rural <- percentrural_by_country$`2019`[percentrural_by_country$`Country Code` == country]
+frac_rural <- country_rural/100 #turn percentage into fraction
+```
+
+The last step in our data prep is to create the vector of normalized population per age group for the specific country. The contact matrices have 5-year age groups with the oldest group begin 80+. Because the age demographics go up to 100+, we need to sum the last five age groups in the demographics vector to ensure compatibility with the contact matrices.
```
rccreswell:
```suggestion
The last step in our data prep is to create the vector of normalized population per age group for the specific country. The contact matrices have 5-year age groups with the oldest group being 80+. Because the age demographics go up to 100+, we need to sum the last five age groups in the demographics vector to ensure compatibility with the contact matrices.
```

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+```{r}
+b = 0.3 # probability of infection
+k = 0.2 # 1/(incubation period in days)
+g = 0.1# 1/(days between infection and recovery)
+m = 0.03 # probability of death, cases-fatality ratio.
+C = 1 # connectedness parameter
+transmission_parameters(model) <- list(b, k, g, m, C)
+```
+
+Next we set the initial conditions of the model, where we specify separately what fraction of the population is infected in the urban community and what fraction is infected in the rural community. Note here that these are fractions of the total population across both communities! Because the populations are normalized, all initial conditions together must sum to one.
```
rccreswell: While it's obvious from the code, you could mention in the text above as well that we are starting the infection in the urban population

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+tend = 80
+time = seq(0, tend, by = 1)
+
+# run simulation
+output <- run(model, time)
+dim(output)
+```
+
+The output from the model can then be plotted such that all populations are plotted separately for both communities
+```{r, fig.height = 4, fig.width = 7}
```
rccreswell: This figure could definitely be bigger for easy visibility

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+  geom_line(aes(color = compartment), size = 1.5) +
+  labs(x = "time (days)", y = "fraction of the population",
+       title = glue("Urban and rural populations")) +
+  theme(legend.position = "bottom", legend.title = element_blank()) +
+  scale_color_manual(values = col)
+
+case_df <- subset(output, compartment == "Incidences_U" | compartment == "Deaths_U" | compartment == "Incidences_Y" | compartment == "Deaths_Y")
+case_df$compartment <- factor(case_df$compartment, levels = c("Incidences_U", "Deaths_U", "Incidences_Y", "Deaths_Y"))
+col <- c("Cases_U" = "grey28", "Deaths_U" = "black", "Cases_Y" = "lightgray", "Deaths_Y" = "darkgray")
+inc_plot <- ggplot(case_df, aes(x = time, y = value, fill = compartment)) +
```
rccreswell, line 187: I like the way you use the same general color for each compartment, but different shades for urban/rural.
However, to my eye S_Y is too light and too close in color to the background---is there a darker light blue?

If the shades end up too similar, another thing that might work is to make all the urban lines dashed, if that's possible in r.

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,268 @@
...
+  labs(x = "time", y = "fraction of the population",
+       title = glue("Urban vs rural model for different levels of connectedness")) +
+  scale_color_viridis(discrete = TRUE)+ 
+  labs(color = "C")
+
+```
+
+The above plot shows that up to approximately $C=0.1$, the value of $C$ has a significant impact on total number of infectious individuals across the communities. Varying the value of $C$ between 0.1 and one, however, barely makes a difference. As both communities are subject to the same parameter values and the average number of contacts that individuals in both communities have don't differ a lot in most countries, this is results is not surprising. Above a certain level of interaction, the few infectious people in one community will quickly interact with individuals from the other community, spreading the infection. At that point, both communities develop their own epidemic under the same parameter values and the mixing between them further helps in synchronizing the waves of the epidemic in both communities. Only for very low levels of interaction does the initial difference in the number of infectious individuals have a lasting effect on epidemic dynamics.
```
rccreswell:
```suggestion
The above plot shows that up to approximately $C=0.1$, the value of $C$ has a significant impact on total number of infectious individuals across the communities. Varying the value of $C$ between 0.1 and one, however, barely makes a difference. As both communities are subject to the same parameter values and the average number of contacts that individuals in both communities have don't differ a lot in most countries, this result is not surprising. Above a certain level of interaction, the few infectious people in one community will quickly interact with individuals from the other community, spreading the infection. At that point, both communities develop their own epidemic under the same parameter values and the mixing between them further helps in synchronizing the waves of the epidemic in both communities. Only for very low levels of interaction does the initial difference in the number of infectious individuals have a lasting effect on epidemic dynamics. 
```

rccreswell (changes requested):
This looks like a very nice implementation and explanation of an interesting model---thanks @solveigvandervegt 
I gave it a first read through and have some suggestions, a few of which might help us get the tests passing

vignettes/SEIRD_rural_urban.Rmd
solveigvandervegt: yes, we do, thanks for the spot :)

DESCRIPTION
solveigvandervegt: probably not, no, I'll take it out!

vignettes/SEIRD_rural_urban.Rmd
solveigvandervegt: Good point, I have added it!

vignettes/SEIRD_rural_urban.Rmd
solveigvandervegt: I have made them bigger, let me know if this is big enough!

vignettes/SEIRD_rural_urban.Rmd
solveigvandervegt: I have chosen a slightly different shade which is hopefully a bit more blue and a bit less gray. Let me know if it is still not right!

solveigvandervegt:
@rcw5890 Thank you for going over this so carefully!!

rccreswell (changes requested):
It's interesting that we are now passing on some operating systems but not on others.

Some of them seem to be complaining that your pull request adds a folder called "Data" when we already have one called "data". Presumably, you can just combine them into "data" and update the file paths.

The rgl error is more annoying. I'm guessing this happens because you are now importing matlib, which suggests rglwidget, which imports rgl (and rgl is apparently very difficult to install on macos). But, do we even need matlib? Your notebook seems to run without it, so I would try removing it as a first attempt.

solveigvandervegt:
@rcw5890 thanks for taking another look! I fixed the matlib issue. I was using a function from that package but I found another way to do it, so the package has now been removed and the associated checks now pass. I am not sure what is going on with the error in ubuntu. I don't have a folder named Data, only data, which for as far as I can tell, is the preferred way of storing data in an R package. I thought that maybe the issue came from the fact that there is a folder named data and a file named data.R, but again, this is all following the guidelines for a proper R package. Since the other os don't complain, it must an Ubuntu issue

rccreswell:
Hmm, the one you have on github does seem to have both Data and data.
This is what I'm looking at:
https://github.com/Como-DTC-Collaboration/como-models/tree/i16-urban-vs-rural-communities/data
https://github.com/Como-DTC-Collaboration/como-models/tree/i16-urban-vs-rural-communities/Data

solveigvandervegt:
hmmmm, that's really weird cause on my local machine, I don't, so I am not sure where the Data folder is coming from!

solveigvandervegt:
For some reason, it appears that once I push to the origin, it is splitting all the data files into two folders. Let me see if I can change some things around

solveigvandervegt:
all fixed :)

vignettes/SEIRD_rural_urban.Rmd
rccreswell: The size looks fine to me now

vignettes/SEIRD_rural_urban.Rmd
rccreswell: I think the one you picked stands out well enough against the gray, thanks

rccreswell:
> all fixed :)

Thanks for making all those changes and getting the tests passing. This is looking great but I did have a couple of other various thoughts, not sure if github is hiding them?

solveigvandervegt:
> > all fixed :)
> 
> Thanks for making all those changes and getting the tests passing. This is looking great but I did have a couple of other various thoughts, not sure if github is hiding them?

Ah, yeah I think I just saw them! I will make the changes tomorrow!

vignettes/SEIRD_rural_urban.Rmd
solveigvandervegt: I cannot tell what the change is here, sorry, could you clarify?

rccreswell: Sorry about that !
My suggestion is "later" -> "latter" in sentence 4

solveigvandervegt: Ah got it, thanks!

R/SEIRD_RU.R
solveigvandervegt: Done!

solveigvandervegt:
I have made the changes you requested. There are some conditions on inputs, but hopefully those are explained sufficiently in the documentation and the vignette. I am not sure what is going on with Windows, sometimes I run the checks and it is fine, and sometimes it is not.

solveigvandervegt:
I think it is ready, but let me know what y'all think!

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,260 @@
...
+tend = 80
+time = seq(0, tend, by = 1)
+
+# run simulation
+output <- run(model, time)
+dim(output)
+```
+
+The output from the model can then be plotted such that all populations are plotted separately for both communities
+```{r, fig.height = 5, fig.width = 10}
```
rccreswell: Some of the text in the figures is getting cut off when I run it. 
Maybe try increasing the width, or switching to nrows=2?

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,260 @@
...
+  geom_bar(stat="identity", position = position_dodge()) +
+  labs(x = "time (days)", y = "fraction of the population") +
+  theme(legend.position = "bottom", legend.title = element_blank(), text = element_text(size = 20)) +
+  scale_fill_discrete(labels = c("Cases_U","Deaths_U","Cases_Y","Deaths_Y"))
+
+grid.arrange(SEIRplot,inc_plot,nrow = 1)
+```
+or such that the populations are summed across the communities and we can observe the dynamics of the susceptible, exposed, infectious, recovered and dead individuals in the whole population.
+
+```{r , fig.height = 5, fig.width = 10}
```
rccreswell: As above, this figure size isn't quite large enough to hold all the legend entries

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,260 @@
...
+```
+
+### The basic reproduction number: R0
+We can compute the basic reproduction number using the next generation matrix method explained by van den Driessche [-@vandenDriessche17]. The basic reproduction for the model can be displayed by simply calling
+```{r}
+print(R0(model))
+```
+
+## The effect of varying $C$
+The connectedness parameter $C$ indicates how connected the two communities are, i.e. how much the individuals in each are expected to interact with each other. At the extreme ends we have the communities completely connected and acting as one big community for $C=1$, and completely disconnected and acting as separate communities for $C=0$. We explore what happens for intermediate values by simulating the model for a range of values of $C$ and plotting the summed the infectious population for each simulation.
```
rccreswell:
```suggestion
The connectedness parameter $C$ indicates how connected the two communities are, i.e. how much the individuals in each are expected to interact with each other. At the extreme ends we have the communities completely connected and acting as one big community for $C=1$, and completely disconnected and acting as separate communities for $C=0$. We explore what happens for intermediate values by simulating the model for a range of values of $C$ and plotting the summed infectious population for each simulation.
```
or,  we could say something like "combined infectious population"

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -0,0 +1,260 @@
...
+    \frac{\text{d}I_U}{\text{d}t} = & \kappa E_U - (\gamma + \mu)I_U ,\\
+    \frac{\text{d}I_Y}{\text{d}t} = & \kappa E_Y - (\gamma + \mu)I_Y,\\
+    \frac{\text{d}R_U}{\text{d}t} = & \gamma I_U,\\
+    \frac{\text{d}R_Y}{\text{d}t} = & \gamma I_Y,\\
+    \frac{\text{d}D_U}{\text{d}t} = & \mu I_U,\\
+    \frac{\text{d}D_Y}{\text{d}t} = & \mu I_Y.\\
+\end{split}
+\end{equation}
+
+\noindent In the above equations, $\phi_U$ is the fraction of the population that lives in an urban environment, $\phi_Y$ is the fraction of the population that lives in a rural environment, $N_U$ is the average number of contacts that an urban individual has in a day (normalized over the whole population), and $N_Y$ is the average number of contacts that a rural individual has in a day (normalized over the whole population), so $(\phi_U N_U + \phi_Y N_Y)$ is the normalized average number of contacts for any individual. $C$ is the connectedness parameter, where $C=1$ means the communities are well-mixed with each other, and $C=0$ means the two communities do not interact at all.
```
rccreswell: I think I see what you are doing here but I still don't see exactly what you mean by "(normalized over the whole population)" and similarly "normalized average number of contacts." Do we even need these statements, or is "average number of contacts" enough?

R/SEIRD_RU.R
```diff
@@ -0,0 +1,642 @@
...
+    #check if input is matrices or numbers, process accordingly
+    if (length(contact_mat$urban) > 1) {
+      # check if age demographics have been provided
+      if(length(country_demog(object)$urban) == 0 || length(country_demog(object)$rural) == 0) {
+        stop("If you wish to use contact matrices for the model, you must provide
+             vectors of the age demographics of the rural and urban communities
+             BEFORE assigning the contact matrices.")
+      }
+    # compute contacts N_U and N_Y from contact matrices
+    # number of urban contacts for an urban individual, normalized over the
```
rccreswell: If you don't mind more of my harping on the "normalized" terminology---is there a reason why we can't just describe this as an average or weighted average?

R/SEIRD_RU.R
```diff
@@ -0,0 +1,642 @@
...
+#' infected individuals (in our case the exposed and infectious compartments in
+#' both communities), and those that do not (the susceptible, recovered and dead).
+#' We only further consider the system of equations consisting of the ODEs 
+#' describing the variables in the first group,
+#' in this case (E_U, E_Y, I_U and I_Y).
+#' The differential equations for the variables in the first group are written
+#' in the form
+#' \deqn{\frac{dx_i}{dt} = F_i - V_i}
+#' where F_i are the terms describing how individuals enter the
+#' compartment and V_i the terms describing how individuals leave.
```
rccreswell: Thanks for adding the extra information here.
I plan to actually check your code, but right now I'm still learning the method. I have a few questions up front---
- The paper says that the derivatives should be evaluated at the disease free equilibrium, but it looks like you are evaluating at the actual initial conditions of the model. Is this allowed?
- The paper you cite says that "V_i(x) is the rate of other transitions between compartment i and other
**infected** compartments." But in most other descriptions of the method, and your implementation, V_i seems to contain transitions to non-infected compartments as well, such as R. Do you have any idea what they mean by this?
- `where F_i are the terms describing how individuals enter the compartment and V_i the terms describing how individuals leave.` I think we can specify that, in general, F_i contains only those individuals entering the compartment that are considered new infections. As for V_i, however my previous question is answered, it seems that it's not restricted to individuals leaving but can contain other transitions.

rccreswell (changes requested):
Thanks for making those changes
There are a few more minor things I noticed

R/SEIRD_RU.R
solveigvandervegt: I think my problem with that is that the populations in the model are all normalized and the units of the entries in the matrix simply aren't "number of contacts" so it seems wrong to note it as such just to simplify things. Since the number of contacts has been divided through by the total population, the numbers in the matrix are much smaller than what you would expect if you were looking at the average number of contacts, so the results we get wouldn't make sense. I understand that I need to phrase it differently cause you don't think it is clear, so I will think about it, but I do think it needs to be noted!

solveigvandervegt: I think with respect to your comment above in the vignette it is different because it is a general explanation of the model without reference to our specific implementation of it.

R/SEIRD_RU.R
solveigvandervegt: In response to your points:
1) Yes, you are correct, we need to be at a disease-free equilibrium. I have fixed this now. Not sure why I used the initial conditions before, probably because they are very, very close to the disease-free equilibrium. Thanks for catching this!
2) I have simply followed the example in the paper. If you want to do it differently, please let me know what maths you want changed and why. I am happy with how it is now.
3) done.

rccreswell (approved):
Thanks I think you've covered everything I saw
I'd be happy for @ben18785 and @IdilCazimoglu to take a look too

solveigvandervegt:
@rcw5890   Great, thank you so much for all your help Richard!!

PR #32, opened 2021-09-01
I20 age structured seird model with NPIs

oscardm20994:
Adds model which simulates non-pharmaceutical interventions to an age structured SEIRD model


hfldai:
### Checklist & log

- [x] Check all docstrings
- [x] + ode functions in docstrings
- [x] Allow arbitrary number of interventions
- [x] Examine time t in run() when time step is not integers
- [x] Vignette (sensitivity; compare with different interventions)
- [x] Transmission param `beta_isolated` - consider renaming? NPI may not only be isolation
- [x] Change model structure: NPI changes social contact matrices by some % of original, depending on different strategies of NPI. Some NPI also changes beta | [Ref](https://www.thelancet.com/journals/lanpub/article/PIIS2468-2667(20)30133-X/fulltext)
- [x] Model renaming: SEIRDAge_interventions -> SEIRDAge_NPI
- [x] Allow different values for each transmission param for each age group
- [x] Include basic reproduction number
- [x] Complete vignette: 1) COVID19-related transmission param values; 2) R0 details; 3) Subtitles; 4) Review for models 5) Logistics & References
- [x] Complete unit test

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,452 @@
...
+
+    # function for RHS of ode system
+    Intervention = GetIdxIntervention(object, times)
+
+    right_hand_side <- function(t, state, parameters) {
+      with(
+        as.list(c(state, parameters)),
+        {
+
+          intv <- Intervention[which(as.integer(t)==times)]
```
ben18785: I'm not quite sure I understand these functions to do with the intervention. In particular, I don't quite see why we are using "times" (which I think is the times at which we report the solution to the ODE) to determine something on the RHS of the ODE? (Which I assume is the index of the particular intervention in place?) I think we would want a function that accepts "t" as an input and returns the index of the intervention currently happening, but I don't see why that should depend on "times"...perhaps I'm missing something?

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,452 @@
...
+    # fetch number of age catagories
+    age <- object@n_age_categories
+
+    # set initial state vector
+    state <- c(S = object@initial_conditions$S0,
+               E = object@initial_conditions$E0,
+               I = object@initial_conditions$I0,
+               R = object@initial_conditions$R0,
+               D = object@initial_conditions$D0,
+               cc = rep(0, age))
```
ben18785: Does "cc" mean those individuals who are cocooning or something else? In any case, we probably want to use capital letters like for the other states?

ben18785 (changes requested):
@hfldai -- looks great. I just had a few questions about the meaning of compartments and the intervention function selector thingy. If you could take a look that'd be great.

vignettes/SEIRD_age_structred_with_interventions.Rmd
```diff
@@ -0,0 +1,136 @@
...
+)
+```
+
+```{r setup}
+library(comomodels)
+library(tidyverse)
+```
+
+## Introduction
+A key aspect of modeling the spread of COVID-19 is the implementation of non-pharmaceutical interventions such as self isolating reducing physical contact through social distancing and shielding vulnerable groups. In this vignette, we implement the effect of a number of such interventions and examine their influence virus dynamics using an age structured SEIRD model.
```
rccreswell:
```suggestion
A key aspect of modeling the spread of COVID-19 is the implementation of non-pharmaceutical interventions such as self isolating reducing physical contact through social distancing and shielding vulnerable groups. In this vignette, we implement the effect of a number of such interventions and examine their influence on virus dynamics using an age structured SEIRD model.
```

rccreswell:
```suggestion
A key aspect of modeling the spread of COVID-19 is the implementation of non-pharmaceutical interventions such as self isolating reducing physical contact through social distancing and shielding vulnerable groups. In this vignette, we implement the effect of a number of such interventions and examine their influence virus transmission using an age structured SEIRD model.
```

vignettes/SEIRD_age_structred_with_interventions.Rmd
```diff
@@ -0,0 +1,136 @@
...
+
+For this vignette we utilise an example case for social contact and subsequent virus dynamics. The contact matrices, $C$ indicate the expected number of contacts an individual of age class $i$ experiences with age class $j$ per day. In this package, we include country-level estimates of contact matrices (Prem et al., 2017; for full details see data documentation). There are actually four types of contact matrix corresponding to daily numbers of contacts at home, work, school and all other settings. Here, we visualise these for India.
```
rccreswell:
```suggestion
For this vignette we utilise an example case for social contact and subsequent virus transmission. The contact matrices, $C$ indicate the expected number of contacts an individual of age class $i$ experiences with age class $j$ per day. In this package, we include country-level estimates of contact matrices (Prem et al., 2017; for full details see data documentation). There are actually four types of contact matrix corresponding to daily numbers of contacts at home, work, school and all other settings. Here, we visualise these for India.
```
I'd avoid this term as people use it to refer to the progression of a disease within a single host

vignettes/SEIRD_age_structred_with_interventions.Rmd
```diff
@@ -0,0 +1,136 @@
...
+
+## Simulation examples 
+
+Now we create a model that includes this age-structured mixing as well as a proportion of infected individuals self isolating.
+Since the contact matrices have 16 age classes, we instantiate a model with this number of age ranges.
+We also give it transmission parameters, populate the age classes with initial conditions and provide it with our overall contact matrix.
+
+We use more demographics loaded from the population data file provided by UN and, for this example, initialise the model
+with a uniform initial infected population of 1% across all age groups. The remaining 99% of the population starts
+as susceptibles. We set the
```
rccreswell:
```suggestion
as susceptibles. We set the ...
```

vignettes/SEIRD_age_structred_with_interventions.Rmd
```diff
@@ -0,0 +1,136 @@
...
+
+We next simulate the model and visualise all the dynamics of each compartment.
+```{r, fig.width=10, fig.height=5}
+res <- run(model, time=seq(0, 100, by = 1))$states
+
+res %>%
+  filter(compartment!="Incidence") %>%
+  ggplot(aes(x=time, y=value, colour=age_range)) +
+  geom_line() +
+  facet_wrap(~compartment)
```
rccreswell: E, I, and D are difficult to see, they need not share the y axis with S and R

vignettes/SEIRD_age_structred_with_interventions.Rmd
```diff
@@ -0,0 +1,136 @@
...
+intervention_parameters(model) <- list(starts=c(10, 25, 40),
+                                      stops=c(15, 30, 50))
+
+
+
+```
+
+We next simulate the model and visualise all the dynamics of each compartment.
+```{r, fig.width=10, fig.height=5}
+res <- run(model, time=seq(0, 100, by = 1))$states
```
rccreswell:
```suggestion
res <- run(model, time=seq(0, 60, by = 1))$states
```
So that the interesting time period is better visible

vignettes/SEIRD_age_structred_with_interventions.Rmd
```diff
@@ -0,0 +1,136 @@
...
+  comment = "#>"
+)
+```
+
+```{r setup}
+library(comomodels)
+library(tidyverse)
+```
+
+## Introduction
```
rccreswell: I would add to the notebook some extra information, including the equations and the fact that we model self isolation and other interventions by allowing a change in beta.

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,452 @@
...
+    
+    trans_params <- list(isolated_frac, beta_isolated, beta_not_isolated, kappa, gamma, mu)
+    
+    # add names to each value
+    names(trans_params) = object@transmission_parameter_names
+    
+    # check format of parameters b, k and g
+    if(length(beta_isolated) != 1 | length(beta_not_isolated) != 1 | length(kappa) != 1 | length(gamma) != 1){
+      stop('The parameter values should be 1-dimensional.')
+    }
```
rccreswell: If an age structured mu or isolated_frac is provided with the wrong number of age categories, the simulate fails rather ungracefully. 
The check in the normal age structured model seems sensible to me (make sure mu has length either 1 or the number of age categories)

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,452 @@
...
+  dt = times[2] - times[1]
+  Intervention <- rep(0, length(times))
+  n_interventions <- length(object@intervention_parameters[["starts"]])
+
+  for(intv in seq(n_interventions)){
+    Intervention[seq(object@intervention_parameters[["starts"]][intv],
+                       object@intervention_parameters[["stops"]][intv],
+                       dt)] = intv
+  }
+  return(Intervention)
```
rccreswell: The current implementation of interventions depends too much on times (which is just where we want to observe the results and should not affect the actual simulation).
The approach taken by the vaccination model https://github.com/Como-DTC-Collaboration/como-models/blob/875e3ac6a6e70d1106dfc569436628f5098d9035/R/SEIRDV.R#L323  may be an improvement

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,452 @@
...
+
+
+#' @describeIn SEIRDAge_interventions Method to simulate output using from model.
+#'
+#' Solves a system to ODEs which form an
+#' age-structured simple SEIR model. The system of equations for the time
+#' evolution of population fractions in Susceptible (S), Exposed (E), Infected
+#' (I), Recovered (R) and Dead (D) groups in a given age group indexed by i is
+#' given by
+
```
rccreswell: Missing equations

rccreswell (changes requested):
Thank you for picking up this PR @hfldai 
Here are my suggestions from a first reading

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,472 @@
...
+#'
+#' @import deSolve
+#' @import glue
+#' @import tidyverse
+#' @import reshape2
+#' @importFrom methods new
+#' 
+#' @export SEIRDAge_interventions
+#' 
+SEIRDAge_interventions <- setClass('SEIRDAge_interventions',
```
ben18785: Can we rename this model `SEIRDNPIAge`? Interventions is vague and could mean, for example, vaccinations. Also can we rename the file and vignette accordingly?

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,472 @@
...
+#'
+#' @export
+setMethod(
+  'initial_conditions<-', 'SEIRDAge_interventions',
+  function(object, value) {
+    S0 = value$S0
+    E0 = value$E0
+    I0 = value$I0
+    R0 = value$R0
+    D0 = value$D0
```
ben18785: Let's just set D0 to always be zero as that's what we've done elsewhere (and I think it makes sense).

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,472 @@
...
+        #print(paste(idx_interv, i))
+        if ((t>=interventions(object)[[i]]$starts) & (t<=interventions(object)[[i]]$stops)) idx_interv = i
+        i = i+1
+      } 
+      # assign to the idx referring to the non-intervention parameter
+      if(idx_interv < 0) idx_interv = object@n_interventions + 1
+      return(idx_interv)
+    }
+    idx_interv_list <- lapply(times, function(t)idx_interv(t))
+    inter <- approxfun(times, idx_interv_list, rule=2) # avoid NA
```
ben18785: This section of the code looks sufficiently complex (and important!) that I think we should test it. Can we break this all out into two separate functions: `idx_interv` (perhaps renaming it so it's very clear what it does) and another that creates the object you've called `inter`? Then add tests for these. Also, can I check, don't we need to be careful about using `times` in case the timesteps are really large?

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,472 @@
...
+          # rate of change
+          #dS <- - S*(beta_interventions * C %*% I  * inter + beta * C %*% I * (1 - inter))
+          dS <- - S*(beta_final * C_final %*% I)
+          #dE <- S*(beta_interventions * C %*% I * inter + beta * C %*% I * (1 - inter)) - kappa * E
+          dE <- S*(beta_final * C_final %*% I) - kappa * E
+          dI <- kappa * E - gamma * I - mu * I
+          dR <- gamma * I
+          dD <- mu * I
+          #dcc <- S*(beta_interventions * C %*% I * inter + beta * C %*% I * (1 - inter))
+          dcc <- S*(beta_final * C_final %*% I)
```
ben18785: Can we remove the commented out equations?

R/SEIRDAge_interventions.R
```diff
@@ -0,0 +1,472 @@
...
+#' @describeIn SEIRDAge_interventions Sets the transmission_parameters for an
+#' age-structured SEIRD model with interventions.
+#' 
+#' Sets transmission_parameters of an
+#' age-structured SEIR model.
+#'
+#' If the transmission parameters provided to are not 1-dimensional an error is
+#' thrown.
+#'
+#' @param value a named list of form list(beta=, kappa=, gamma=, mu=)
```
ben18785: This isn't correct as it stands: beta_interventions, for example, is missing. Also, I think that beta vs beta_interventions needs explaining.

R/SEIRDAge_interventions.R
hfldai: (Now solved by setting up a function)

R/SEIRDAge_interventions.R
hfldai: "cc" indicates the cumulative number of cases over time. It is "C" in the basic model, but in models with contact matrices, we name the contact matrices as "C" in ODEs and "cc" for cumulative cases instead. I simply took the same namings from other age-structured models.

Do we want the names be changed?

vignettes/SEIRD_age_structred_with_interventions.Rmd
hfldai: (Notebook modified with information added)

R/SEIRDAge_interventions.R
hfldai: (Solved by modifying model structure and setting up a function taking t as variables in dealing with interventions)

R/SEIRDAge_interventions.R
hfldai: (added)

ben18785 (approved):
Looks good. Thanks @hfldai

PR #37, opened 2021-09-29
I34 moving shared generics into R/generics.R

hfldai:
Moving generics which are shared across more than one class into [R/generics.R](https://github.com/Como-DTC-Collaboration/como-models/blob/i34-generics-integration/R/generics.R), as requested in issue #21.


DESCRIPTION
```diff
@@ -33,3 +33,10 @@ Imports:
 VignetteBuilder: knitr
 Depends: 
     R (>= 2.10)
+Collate:
```
ben18785, line 36: Ah, I was wondering how to do this so that generics was processed first -- nice!

ben18785 (approved):

PR #38, opened 2021-11-11
SEIRD vignette following Lisa's comments

### LINKED ISSUE #30: Lisa's comments from Como-model presentations
SEIRD:
- we want to include more text about the conditions that this might be appropriate for: specifically about natural birth and death
- need to remove S0 from the R0 equation
- rename R(0) as something else
- proportion infected and recovered opposed to number of infected and recovered
- express incidence as cases per 1000 of the population

SEIRD with age structure:
- again add a caveat about natural birth and death; also mention that it would be easy to include the UN data on demographics which would be important if we wanted to run for longer
- mention also that we would need to include aging in the model if we were running it for longer
- put units on the axes
- shouldn't the SEIRD with age be named SEIRDAge opposed to SEIRAge
- we should probably go into infection fatality ratios for this model

Uncertainty in contact matrix:
- add in a comment about contact matrices being representative of the whole country; mention about the granularity of data

Hui Jia: optimisation:
- It's worth indicating that we are not including reporting fraction
- Make the axis limits the same for the profile likelihood plots

Numerical solvers:
- Might be worth talking about how variable step size solvers work

Ioana: SEIRDV
- Make axis fonts larger
- Scenario analysis: try starting vaccination at different points in the epidemic
- Mention in the vignette that we may be in fact vaccinating E, I and Rs- Produce some graphs with interventions on / off

Idil: SEIRD with birth and natural death and the non-pharmaceutical interventions
- Replenishments for Covid happen at a faster rate than how births and deaths influence things: important to caveat this in the vignette
- Again would be good to do an intervention analysis, I think: turn on vs off

Liangti: GUI
- Add a disclaimer about "don't blame me if you use this to make a decision and it's bad". Have a look at the Como app for this.
- Change "State" to "Model Compartments"
- Lisa to share a "building up complexity" pres for malaria

FarmHJ:
- Included more description on the conditions when the model can be used
- Renamed S_0, E_0, I_0, R_0... to S_{init}, E_{init}, I_{init}, R_{init} in the vignette to avoid confusion of R_0 as initial value of R and reproductive number. 
- Expressed incidences as cases per 1000 of the population instead of fraction of population


vignettes/SEIRD.Rmd
```diff
@@ -15,75 +15,188 @@ knitr::opts_chunk$set(
...
+
+Note that $\beta$ is a rate (per unit time) at which a susceptible individual becomes infected on contact with an infected individual. It is not a probability as it is sometimes mistakenly labelled and so $\beta > 1$ is possible.
+
+The full SEIRD model consists of four ODEs that a coupled together: the first of these governs the behaviour of susceptible individuals, which has already been introduced. The next equation governs the growth of the exposed compartment:
```
rccreswell:
```suggestion
The full SEIRD model consists of four ODEs that are coupled together: the first of these governs the behaviour of susceptible individuals, which has already been introduced. The next equation governs the growth of the exposed compartment:
```

vignettes/SEIRD.Rmd
```diff
@@ -15,75 +15,188 @@ knitr::opts_chunk$set(
...
+  ggplot(aes(x=t, y=E, colour=as.factor(kappa))) +
+  geom_line() +
+  scale_color_brewer("kappa", palette = "Dark2") +
+  xlab("time (days)")
+```
+An alternative way to view these exponential decay curves is as probability distributions governing the time taken for an exposed individual to become infectious. This period is referred to as the latency period. Whilst rigorous proof of this requires considering the equivalent stochastic version of this model, the mean from the deterministic model matches the mean of the stochastic equivalent. As such, the waiting time for an individual to transition from exposed to infectious classes is governed by an exponential distribution. Using repeatedly sampled PCR test data from close contacts of confirmed cases in mainland China, a recent study estimated the latency period of COVID-19 as 5.51 days and a log-normal probability distribution was found to represent that waiting period (Xi et al., Estimating the Latent Period of Coronavirus Disease 2019, 2021). Below we compare the waiting period for the log-normal distribution fit in this study with an exponential distribution with the same latency period.
```
rccreswell: I think this paragraph needs 1--2 sentences extra explanation---some readers may be confused that we are suddenly talking about probability distributions and a mean in the deterministic model.

vignettes/SEIRD.Rmd
```diff
@@ -15,75 +15,188 @@ knitr::opts_chunk$set(
...
+
+## Steady states
+We can now consider whether there are equilibrium states of the system: values of the state variables where, unless perturbed, the system will remain where it is. To do so, we set the right-hand sides of the ODEs to zero. Taking the equation for the 
+susceptibles:
+
+$$\frac{\text{d}S}{\text{d}t} = -\beta S I = 0.$$
+
+This only has solutions if either $S=0$, in which case there are no individuals left to infect; or $I=0$, in which case there are no infectious individuals: in other words, an infection has not been imported into the population. Considering all four ODEs, $I=0$ implies that $\frac{\text{d}R}{\text{d}t} = 0$ and $\frac{\text{d}D}{\text{d}t} = 0$. Additionally, so long as there are no exposed individuals, $\frac{\text{d}E}{\text{d}t}= 0$ and $\frac{\text{d}I}{\text{d}t}= 0$ and 
+So the equilbria are given by $S=S'$, $E=0$, and $I=0$ which corresponds to a situation when there are no infected or infectious individuals in the population. In other words, in this model, there is no endemic equilbrium when the infection remains present throughout the population; only epidemic dynamics are possible.
```
rccreswell:
```suggestion
So the equilibria are given by $S=S'$, $E=0$, and $I=0$ which corresponds to a situation when there are no infected or infectious individuals in the population. In other words, in this model, there is no endemic equilibrium when the infection remains present throughout the population; only epidemic dynamics are possible.
```

vignettes/SEIRD.Rmd
```diff
@@ -158,35 +278,62 @@ for(i in seq_along(beta_vals)) {
...
+  scale_color_viridis_c() +
+  xlab("time (days)") +
+  ylab("Infectious proportion")
 ```
 
-Why do some epidemics die out quickly but others don't? This is can be explored through the basic reproductive number, $R_0$, which is the expected number of people infected by a
-single infectious individual in a population where everyone is susceptible (i.e $S_0 \approx 1$). Note that this basic reproductive number is unrelated to the initial condition for the recovered population! If $R_0>1$, an infectious individual infects more than one other person and the epidemic initially grows exponentially; if $R_0<1$, an infectious individual infects less than one other person and the epidemic dies out quickly. In an SEIRD model with normalized populations, the reproductive number is given by:
+Why do some of the simulations produce epidemic waves whereas others fail to do so?
+
+This is can be explored through the basic reproductive number, $R_0$, which is the expected number of people infected by a single infectious individual in a population where everyone is susceptible. (Note that this basic reproductive numberis unrelated to the initial condition for the recovered population.)
```
rccreswell:
```suggestion
This is can be explored through the basic reproductive number, $R_0$, which is the expected number of people infected by a single infectious individual in a population where everyone is susceptible. (Note that this basic reproductive number is unrelated to the initial condition for the recovered population.) 
```

rccreswell (changes requested):

vignettes/SEIRD.Rmd
ben18785: Thanks Richard. I've changed this now.

ben18785:
Closing as is now being covered in #46

PR #41, opened 2021-11-14
I26 vaccination age struct

I-Bouros:
## Issue
- [x] Which issue does your pull request address? If none of the existing issues are applicable, you can write a new issue.

## Testing, documentation, and checks
- [x] If your pull request introduces new code, you should write automated tests (explained [here](https://r-pkgs.org/tests.html)).
- [x] Check that all tests pass using `devtools::check()`.
- [x] Check compliance with the preferred R style using `lintr::lint_package()`.
- [x] Build documentation using `devtools::document()`.
- [x] If you changed the readme, run `knitr::knit("README.Rmd")`.

## Pull request
- [x] Assign at least one reviewer.


R/SEIRDVAge.R
```diff
@@ -0,0 +1,506 @@
...
+#'
+#' Intervention parameters have same size. A tanh function is used to smooth interventions during simulation. This class is designed for interventions
+#' which last several days at least and have several days between them; interventions involving rapid fluctuations may be distorted.
+#'
+#' @param object an object of the class SEIRDVAge
+#' @param value (list) list of intervention parameters: starts, stops and
+#'              coverages. Coverages can be either a vector, each value representing
+#'              the level of intervention for the corresponding pair of start-stop
+#'              time points, or a list of vectors, each element of the list representing
+#'              the same timeline corresponding to each age-group
```
rccreswell: The format of age structured parameters seems very complicated. Why not allow age structured parameters to be a list of the parameters for a single age group, such that
```
all_int_params = vector(mode="list", length=n_ages)
for(i in 1:n_ages){
  all_int_params[[i]] = list(starts=c(...),
                             stops=c(...),
                             coverages=c(...))
}
```
would be the correct input?
In any case, I would expand the explanation of the correct format, both here and in the notebook.

R/SEIRDVAge.R
```diff
@@ -0,0 +1,506 @@
...
+#' (I), Recovered (R), Vaccinated (V) and Dead (D) groups in a given age group indexed by i is 
+#' given by
+#'
+#' \deqn{\frac{dS_i(t)}{dt} = - \beta S_i(t) \Sigma_{j}C_{ij} I_j(t) - \nu Inter(t) S_i(t) + \delta_V V_i(t) + \delta_R R_i(t) + \delta_VR VR_i(t)}
+#' \deqn{\frac{dE_i(t)}{dt} = \beta S_i(t) \Sigma_{j}C_{ij} I_j(t) - \kappa E_i(t)}
+#' \deqn{\frac{dI_i(t)}{dt} = \kappa E_i(t) - \gamma I_i(t) - \mu I_i(t)}
+#' \deqn{\frac{dR_i(t)}{dt} = \gamma I_i(t) -  \delta_R R_i(t) - \nu Inter(t) R_i(t)}
+#' \deqn{\frac{dV(t)}{dt} = \nu Inter(t) S_i(t) - \delta_V V_i(t)}
+#' \deqn{\frac{dVR(t)}{dt} = \nu Inter(t) R_i(t) - \delta_VR VR_i(t)}
+#' \deqn{\frac{dC(t)}{dt} = \ beta S_i(t) \Sigma_{j}C_{ij} I_j(t)}
```
rccreswell:
```suggestion
#' \deqn{\frac{dC(t)}{dt} = \beta S_i(t) \Sigma_{j}C_{ij} I_j(t)}
```

R/SEIRDVAge.R
```diff
@@ -0,0 +1,506 @@
...
+    
+    if(is.numeric(intervention_parameters(object)$coverages)){
+      int_parms <- 
+        InterventionParameters(
+          start=intervention_parameters(object)$starts,
+          stop=intervention_parameters(object)$stops,
+          coverage= intervention_parameters(object)$coverages)
+      inter_prot <- intervention_protocol(int_parms, sim_parms, 1) %>%
+        filter(time %in% times) %>%
+        .$coverage
```
rccreswell: Is the filtering necessary?
```
    inter_prot <- intervention_protocol(int_parms, sim_parms)
    intervention <- approxfun(inter_prot$time, inter_prot$coverage, rule=2)
```

R/SEIRDVAge.R
```diff
@@ -0,0 +1,506 @@
...
+      inter_prot <- vector(mode = "list", length = n_age)
+      for(age_cov in intervention_parameters(object)$coverages){
+        int_parms <- 
+          InterventionParameters(
+            start=intervention_parameters(object)$starts,
+            stop=intervention_parameters(object)$stops,
+            coverage= age_cov)
+        inter_prot[[i]] <- intervention_protocol(int_parms, sim_parms, 1) %>%
+          filter(time %in% times) %>%
+          .$coverage
```
rccreswell: Similarly, I think we can do away with the filter and just send inter_prot.time and inter_prot.coverage to approxfun

R/SEIRDVAge.R
```diff
@@ -0,0 +1,506 @@
...
+    intervention <- function(t){
+      if(is.numeric(inter_prot)){
+        interv_func <- approxfun(times, inter_prot, rule=2)(t)
+      }
+      else{
+        interv_func <- numeric(n_age)
+        for(i in 1:n_age){
+          interv_func[i] <- approxfun(times, inter_prot[[i]], rule=2)(t)
+        }
+      }
```
rccreswell: I don't like the way it has to perform the interpolation again each time the function `intervention` is called.
Can we do the approxfun outside, and have this function just use previously saved approxfuns

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,464 @@
...
+```
+The added intervention feature allows for variable levels of vaccination in a population.
+In most cases, a vaccine is not available or considered necessary right from the
+start of an epidemic so there is a delay in vaccination. Similarly, vaccination
+programs may not run at the same rate all the time, so such fluctuations need to
+be taken into account in our modeling. Vaccination levels and start times may differ between
+different age groups. In this instantiation of the model, vaccination of the each age group occurs
+at a different stage
+
+Below we can see a plot of the a claassical level of interventions that occur at each time step.
```
rccreswell:
```suggestion
Below we can see a plot of the a classical level of interventions that occur at each time step.
```

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,464 @@
...
+                                    coverage=c(1, 1, 1))
+
+sim_parms <- SimulationParameters(start=0, stop=200, tstep = 0.1)
+
+intervention_protocol(int_parms, sim_parms, 1) %>%
+  ggplot(aes(x=time, y=coverage)) +
+  geom_line() +
+  scale_color_brewer(palette = "Dark2") +
+  labs(x = "time (days)", y = "level of intervention") +
+  theme(text = element_text(size = 20))
```
rccreswell: It would be better to here plot the actual age structured intervention, either with different colors for each age group or one panel for each group's intervention

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,464 @@
...
+  theme(text = element_text(size=20),
+        axis.text.x = element_text(angle = 90, vjust = 0.5, hjust=1)) +
+  geom_tile() +
+  scale_fill_viridis_c() +
+  facet_wrap(~type)
+```
+
+We can see that the majority of contacts fall between younger individuals, occurring at home or "other" locations. At home, however, there is more substantial intergenerational mixing than in other settings.
+
+The model we introduce here does not compartmentalise individuals according to location, so we make an overall contact matrix by summming these together.
```
rccreswell:
```suggestion
The model we introduce here does not compartmentalise individuals according to location, so we make an overall contact matrix by summing these together.
```

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,464 @@
...
+  facet_wrap(~delta_VR) +
+  geom_line() +
+  scale_colour_viridis_d() +
+  labs(x = "time (days)", y = "fraction of the population \n that died per day") +
+  theme(text = element_text(size = 20)) 
+
+```
+As we can see from the plot above, larger rates of immunity decay $\delta_VR$ in vaccinated recovered
+individuals make the second wave of the number of deaths in the epidemic occur much closer to the first
+peak. Also, the earlier the second wave occurs, the higher its applitude will be.
```
rccreswell:
```suggestion
peak. Also, the earlier the second wave occurs, the higher its amplitude will be.
```

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,464 @@
...
+  filter(compartment=="Deaths") %>% 
+  ggplot(aes(x=time, y=value, color = age_range)) +
+  facet_wrap(~nu_2530) +
+  geom_line() +
+  scale_colour_viridis_d() +
+  labs(x = "time (days)", y = "fraction of the population \n that died per day") +
+  theme(text = element_text(size = 20))
+
+```
+As we can see from the plot above, larger rates of maximum vaccination $\nu_{25-35}$ make the second wave of the number of deaths in the epidemic occur later compared to the first peak. For larger $\nu_{25-35}$, we that the hight of the second peak is larger.
```
rccreswell:
```suggestion
As we can see from the plot above, larger rates of maximum vaccination $\nu_{25-35}$ make the second wave of the number of deaths in the epidemic occur later compared to the first peak. For larger $\nu_{25-35}$, we see that the height of the second peak is larger.
```

rccreswell (changes requested):
If it's not too difficult to figure out reasonable values of nu and coverage for this---Could you add to the notebook a pair of simulations comparing two interventions, both with approximately the same total rate of vaccination, but in one case applied equally to all age groups, and in the other applied to each age group in succession starting with the oldest. This would make most sense with an age structured death rate similar to the contact matrix notebook https://github.com/Como-DTC-Collaboration/como-models-math-biosci/blob/e1ffcc491eaec2a7d6ce1b5a0d99ee548f1a49af/contact_matrix/SEIRD_contact_matrix.Rmd#L194

You would show that the second simulation has fewer total deaths, even though the population vaccination rates are comparable. This is not a surprising result, but it is a highly relevant one, and justifies the complexity of the model.

I-Bouros:
> If it's not too difficult to figure out reasonable values of nu and coverage for this---Could you add to the notebook a pair of simulations comparing two interventions, both with approximately the same total rate of vaccination, but in one case applied equally to all age groups, and in the other applied to each age group in succession starting with the oldest. This would make most sense with an age structured death rate similar to the contact matrix notebook https://github.com/Como-DTC-Collaboration/como-models-math-biosci/blob/e1ffcc491eaec2a7d6ce1b5a0d99ee548f1a49af/contact_matrix/SEIRD_contact_matrix.Rmd#L194
> 
> You would show that the second simulation has fewer total deaths, even though the population vaccination rates are comparable. This is not a surprising result, but it is a highly relevant one, and justifies the complexity of the model.

I had a discussion with Ben yesterday and while he agrees with the idea of having a time-dependent nu parameter as we have discussed before, it might be better to postpone that idea for a bit.

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,554 @@
...
+Exposed individuals become infectious at rate $\kappa$. $1/ \kappa$ is
+the incubation time of the virus, i.e. how many days after exposure the individual
+becomes infectious to others. Infectious individuals can either recover or die
+because of the disease. Infectious individuals recover in $1/ \gamma$
+days, thus individuals recover at rate $\gamma$. Similarly, infectious
+individuals die due to the infection after $1/ \mu$ days, making the death rate $\mu$.
+Susceptible individuals are vaccinated at a maximum rate $\nu$. Among those who are immune
+(recovered and vaccinated), people lose their immunity against the virus at different
+rates depending on what type of immunity they acquired: $\delta_V$ for those vaccinated,
+$\delta_R$ for those previously infected and $\delta_VR$ for those previously infected that
```
rccreswell:
```suggestion
$\delta_R$ for those previously infected and $\delta_{VR}$ for those previously infected that
```

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,554 @@
...
+  as.matrix()
+
+# create model
+my_model <- SEIRDVAge(n_age_categories = n_ages,
+                 contact_matrix = c_combined_wider,
+                 age_ranges = as.list(age_names))
+```
+
+Next, we set the parameter values for $\beta$, $\kappa$, $\gamma$, $\mu$, $\nu$,
+$\delta_V$, $\delta_R$ and $\delta_{VR}$:
```
rccreswell: Mention that the mu is age structured, and cite the source it came from (Verity)

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,554 @@
...
+  facet_wrap(~compartment, scales="free") +
+  scale_colour_viridis_d() +
+  labs(x = "time (days)", y = "fraction of the population") +
+  theme(text = element_text(size = 20)) 
+
+```
+
+## Sensitivity analysis
+In this section, we investigate the sensitivity of the number of deaths to the loss-of-immunity rate
+in recovered individuals who have also been vaccinated, $\delta_{VR}$. We run the model for varying values of $\delta_VR$ and plot the number of deaths per day over time.
```
rccreswell:
```suggestion
in recovered individuals who have also been vaccinated, $\delta_{VR}$. We run the model for varying values of $\delta_{VR}$ and plot the number of deaths per day over time.
```

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,554 @@
...
+```
+The added intervention feature allows for variable levels of vaccination in a population.
+In most cases, a vaccine is not available or considered necessary right from the
+start of an epidemic so there is a delay in vaccination. Similarly, vaccination
+programs may not run at the same rate all the time, so such fluctuations need to
+be taken into account in our modeling. Vaccination levels and start times may differ between
+different age groups. In this instantiation of the model, vaccination of the each age group occurs
+at a different stage.
+
+Below we can see a plot of the a classical level of interventions that occur at each time step.
```
rccreswell, line 255: Can you add the age groups to the figure somewhere

vignettes/SEIRDVAge.Rmd
```diff
@@ -0,0 +1,554 @@
...
+  filter(compartment=="Deaths") %>% 
+  ggplot(aes(x=time, y=value, color = age_range)) +
+  facet_wrap(~interv_type) +
+  geom_line() +
+  scale_colour_viridis_d() +
+  labs(x = "time (days)", y = "fraction of the population \n that died per day") +
+  theme(text = element_text(size = 20)) 
+
+```
+In the case of a `common` vaccination protocol for all age groups, not only is the first peak of the number of deaths higher than in the case of age-specific interventions, but the second peak tailing the first is also more pronounced and lasts longer and the third wave occurs sooner and is also higher in amplitude then for the `distinct` interventions case. This suggests that a age-dependent policy for vaccination proves to be more effective in reducing the daily number of deaths throughout the epidemic than the one-method-fits-all approach.
```
rccreswell:
```suggestion
In the case of a `common` vaccination protocol for all age groups, not only is the first peak of the number of deaths higher than in the case of age-specific interventions, but the second peak tailing the first is also more pronounced and lasts longer and the third wave occurs sooner and is also higher in amplitude then for the `distinct` interventions case. 
```
Given that we've done nothing to check the accuracy of parameter values (except mu), I'd avoid making such a strong claim.

R/SEIRDVAge.R
```diff
@@ -0,0 +1,497 @@
...
+    }
+    
+    inter <- function(t){
+      interven <- vector("numeric", length = length(interventions(object)))
+      for(i in 1:length(interventions(object))){
+        interven[i] <- approxfun(inter_prot[[i]]$time,
+                                 inter_prot[[i]]$coverage, rule=2)(t)
+      }
+      return(interven)
+    }
```
rccreswell:
```suggestion
    for(i in 1:length(interventions(object))){
      int_parms <-
        InterventionParameters(
          start=interventions(object)[[i]]$starts,
          stop=interventions(object)[[i]]$stops,
          coverage= interventions(object)[[i]]$coverages)
      prot <- intervention_protocol(int_parms, sim_parms, 1)
      inter_prot[[i]] <- approxfun(prot$time,
                                   prot$coverage, rule=2)
    }

    inter <- function(t){
      interven <- vector("numeric", length = length(interventions(object)))
      for(i in 1:length(interventions(object))){
        interven[i] <- inter_prot[[i]](t)
      }
      return(interven)
    }
```
This speeds up the simulations by 4--5 times

rccreswell (changes requested):
Thanks @I-Bouros here are the final things I noticed

rccreswell (approved):

Issue #14, opened 2021-04-22
Add SEIRD model with waning immunity


ben18785:
@DSeiferth and @hfldai pointed me to the Como package where they allow recovered individuals to become susceptible again. In the meeting, we decided to implement a basic SEIRD model where R -> S at a certain rate. We think this'll result in an endemic equilibrium and (possibly) a number of epidemic waves.

hfldai:
Here's the example plot showing the effects when including R -> S:
* R -> S, transmission rate = 0.01
![SEIaImIsRD_omega-0 01](https://user-images.githubusercontent.com/44946606/116530992-9323d980-a8d6-11eb-90a2-c94dd2fca54a.png)
* No transmission from R -> S
![SEIaImIsRD_omega-0](https://user-images.githubusercontent.com/44946606/116531278-e8f88180-a8d6-11eb-8a93-26faecee5da2.png)

ben18785:
Ha! That's really interesting. And, I guess, makes sense in terms of
oscillatory behaviour since S -> I (and a reduction in S) then, later, I ->
R -> S (leading to an increase).

ben18785:
This is in there via #53

PR #53, opened 2021-12-31
Births, natural deaths and waning natural immunity

ben18785:
Adds a model and vignette that allows: births, natural deaths and waning natural immunity (i.e. protection from reinfection due to a previous infection). The vignette shows how epidemic waves occur and gives an order of magnitude estimate for the case counts that would be reached at endemic equilibrium for COVID-19.

Addresses #14.


solveigvandervegt:
Looks really good! I fixed some typos and grammar issues, hope that's ok. The only thing that I have some reservations about is the computation of beta. I know it comes from the other vignette but since you already mention that in the text, I would be tempted to just set the values here and not derive it again. It's quite a confusing computation out of context and if people want to know why these values are used, they know where to look.

solveigvandervegt:
Oh also, the last situation in the references list is missing an author?

solveigvandervegt:
And is it correct that there are no tests for this class? I cannot see them in the test that folder

ben18785:
Thanks @solveigvandervegt -- you're spot on re: tests. I'll add some now. I'll also address your other comments.

ben18785:
Thanks @solveigvandervegt I've added tests. Re: R_0, I actually quite favour having it in there (as a more pedagogical aspect of the vignette). I've added a little more text that explains both the relevance of the calculation (for the vignette itself) and the calculation. Anyhow, see what you think!

solveigvandervegt:
Hey Ben, yeah sorry you are totally right re: R0 and beta, I read it wrong. I think my point is actually about how gamma and mu are computed. The zeta and ifr come a bit out of nowhere, though I know this is addressed in the other vignette. Sorry for the poorly written comment before!

ben18785:
Cool no worries -- I've now added an explicit reference to the other vignette. I actually think that this will partly be addressed by #67 so perhaps we can leave this as is for now and address later?

solveigvandervegt:
Yeah sure, sounds good!

Issue #51, opened 2021-12-27
Make parameterisation of SEIRD specific to COVID-19


ben18785:
Currently parameters are arbitrary.

ben18785:
This is in there via #54

PR #54, opened 2022-01-02
SEIRD parameterisation: made specific to COVID-19

ben18785:
Addresses #51


vignettes/SEIRD.Rmd
```diff
@@ -201,39 +115,75 @@ $$\frac{\text{d}S}{\text{d}t} = -\beta S I = 0.$$
...
+
+Important though is that both approaches use data to determine parameters. And for practically all infectious diseases modelled, the available data do not pin down a model's parameters to a single set. It is thus important to determine whether the conclusions of modelling -- whether a new viral variant will lead a country's healthcare infrastructure to be overrun, for instance -- are sensitive to the parameter choices assumed.
```
I-Bouros:
```suggestion
However, what is most important is that both approaches use data to determine parameters. For practically all infectious diseases modelled, the available data do not pin down a model's parameters to a single set. Thus, it is important to determine whether the conclusions of modelling -- whether a new viral variant will lead a country's healthcare infrastructure to be overrun, for instance -- are sensitive to the parameter choices. 
```

vignettes/SEIRD.Rmd
```diff
@@ -201,39 +115,75 @@ $$\frac{\text{d}S}{\text{d}t} = -\beta S I = 0.$$
...
+
+Here, we follow the first approach to parameterising the SEIRD model, and show later how a sensitivity analysis can be performed.
+
+The SEIRD model has four transmission parameters: $\beta$, $\kappa$, $\gamma$ and $\mu$. We choose $\kappa=1/5.5$, implicitly assuming a latency period of 5.5 days. We assume that 0.66\% of individuals infected with the virus die [@verity2020estimates]: this is termed an infection fatality ratio (IFR). Changing either $\gamma$ or $\mu$ affects both the rate at which individuals recover or die and the proportions which recover or die. As such, the implications of the individual parameter values are hard to intuit without recourse to the other. Here, we choose a different parameterisation that makes it simpler to set these parameters to appropriate values:
```
I-Bouros:
```suggestion
The SEIRD model has four transmission parameters: $\beta$, $\kappa$, $\gamma$ and $\mu$. We choose $\kappa=1/5.5$, implicitly assuming a latency period of 5.5 days. We assume that 0.66\% of individuals infected with the virus die [@verity2020estimates]: this is termed an infection fatality ratio (IFR). Changing either $\gamma$ or $\mu$ affects both the rate at which individuals recover or die and the proportions of the population which recover or die. As such, the implications of the individual parameter values are difficult to divine without recourse to the other. Here, we choose a different parameterisation that makes it simpler to set these parameters to appropriate values:
```

vignettes/SEIRD.Rmd
```diff
@@ -201,39 +115,75 @@ $$\frac{\text{d}S}{\text{d}t} = -\beta S I = 0.$$
...
+\end{equation}
+
+We assume an average duration of infectiousness of 2 days, which indicates that, after 10 days, <1\% would be infectious -- in line with UK government policy that assumes individuals cannot pass on infection after 10 days [@stayathome2021]. As such, we set $\zeta = 1/2$.
+
+The remaining parameter, $\beta$, is set to a value to produce the same basic reproduction number as was assumed in the influential "report 9" from Imperial College [@ferguson2020report]: $R_0=2.4$ (see explanation of $R_0$ below).
+
 ```{r}
-params <- list(beta=1, kappa=0.9, gamma=0.5, mu=0.1)
+kappa <- 1 / 5.5
+IFR <- 0.0066
```
I-Bouros:
```suggestion
load(file = "../data/infection_fatality_ratio.rda")

IFR <- infection_fatality_ratio$overall
```

I-Bouros (changes requested):
We only need to make sure to feed the changes made to main into this branch before merging.

vignettes/SEIRD.Rmd
ben18785: I've taken part of this and kept the other bit as was -- hope ok!

vignettes/SEIRD.Rmd
ben18785: This now isn't needed at all with Solveig's covid_transmission_parameters function, which is great :)

vignettes/SEIRD.Rmd
ben18785: Sorry I prefer mine here!

PR #69, opened 2022-01-06
added covid parameter function

solveigvandervegt:
Add function to be used in vignettes that provides values for kappa, gamma, mu and R0 given a specific variant and whether or not the model is age-structured.


solveigvandervegt:
Thanks for writing the tests! If there is other things you want to change, let me know and I would be happy to take a look at it tomorrow!

PR #71, opened 2022-01-07
update vignettes with covid parameter function

solveigvandervegt:
So far, I have updated all non-age-structured vignettes with new parameter function, except for the vaccine one.

### DIFF
--- a/vignettes/SEIRD_rural_urban.Rmd
+++ b/vignettes/SEIRD_rural_urban.Rmd
@@ -123,13 +123,17 @@ model <- SEIRD_RU()
 
 followed by setting the parameter values of the model that are not determined by the demographics or contact matrices.
 ```{r}
-b = 0.3 # probability of infection
-k = 0.2 # 1/(incubation period in days)
-g = 0.1# 1/(days between infection and recovery)
-m = 0.03 # probability of death, cases-fatality ratio.
+parameters <- covid_transmission_parameters() # base variant
+print(parameters)
+
+kappa <- parameters$kappa
+gamma <- parameters$gamma
+mu <- parameters$mu
+R0_target <- parameters$R0
+
 C = 1 # connectedness parameter
-transmission_parameters(model) <- list(b, k, g, m, C)
 ```
+Setting the value of $\beta$ in this model is a bit more complex, so we will come back to that in a little bit.
 
 Next we set the initial conditions of the model, where we specify separately what fraction of the population is infected in the urban community and what fraction is infected in the rural community. Note here that these are fractions of the total population across both communities! Because the populations are normalized, all initial conditions together must sum to one. In this example, we seed the epidemic in the urban community.
 ```{r}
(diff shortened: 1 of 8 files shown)


vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -121,140 +119,198 @@ We start by creating an instance of the model class,
...
+
+\begin{equation}
+I^{\text{mixed}}(t) = I(0) \exp(\beta t).
+\end{equation}
+
+Instead, now we take the above population and divide it up into two subpopulations of equal size and assume that each has half the starting number of infected individuals. Assume that each of these communities are epidemic islands and are isolated from one another. Then, the early growth rate of the epidemic in each will also be exponential, albeit with half the rate parameters:
+
+\begin{equation}
+I_j(t) = I(0)/2 \exp(\beta t/2),
+\end{equation}
```
solveigvandervegt, line 295: I think you need to motivate this a bit more, at least I don't follow your reasoning why both beta and I(0) should be divide by two. If this comes from your simplified version of dI/dt, then I think it might be good to start there and do the integration again. Intuitively, I can see why you would divide I(0), but beta honestly isn't obvious to me as it is an inherent property of the virus and doesn't have anything to do with population size.

vignettes/SEIRD_rural_urban.Rmd
```diff
@@ -121,140 +119,198 @@ We start by creating an instance of the model class,
...
+The above analysis shows that a population with low connectivity between subpopulations -- countries, regions, cities and so on -- can substantially slow the transit of infections when infections are seeded non-uniformly. This hints at the importance of travel restrictions, particularly during the early phase of epidemics when infections have not had the opportunity to spread across borders.
+
+## The impact of a fragmented population on rates of transmission
+Aside from reducing flow between subpopulations, fragmenting a population into smaller groups also slows spread for another reason: there are fewer contacts to infect. We now use a straightforward mathematical argument to illustrate this.
+
+For simplicity, suppose that the latent period is very short, so that we can neglect it. Also suppose that we consider also a fully mixed population where everyone interacts with everyone else. Our `SEIRD` model then becomes an `SIR` model, with the infectious compartment dynamics determined by:
```
solveigvandervegt, line 279: this sentence has also twice

vignettes/SEIRD_rural_urban.Rmd
solveigvandervegt: Also, and this might be wrong, but I feel like with split populations, initially, when the population is largely susceptible, is exactly the time when it shouldn't matter that you have two communities because spread of the disease won't yet be limited by the number of susceptible individuals. That point comes later when a good number of people have been infected, and then it depends on how much mixing you have with the other population and whether or not that "supplies" enough susceptibles for the epidemic to grow as quickly as it would in one, well-mixed population

solveigvandervegt: If anything, for two very weakly connected populations, the peak in the initially infected populations might be reach before the peak of the well-mixed model because there are less people to infect but the peak in the second population will be much later. I think the point to make there is that the peak number of infections in the two communities model, although you will have two instead of one, will be much lower than in a well-mixed model, which is good when thinking about hospital capacity, etc. (I think that is what I write originally in my explanation of this phenomenon).

ben18785: Thanks @solveigvandervegt So, it's not beta that halves; it's the susceptible population size that does, resulting in this factor of 2:

dIdt = beta S I

If we start with S=1 this becomes

dIdt = beta I

If we split the population in two then each evolves as:

dIdt = beta / 2 I

 You're right though, I need to explicitly show this as it's not clear.

Re: "it shouldn't matter that you have two communities because spread of the disease won't yet be limited by the number of susceptible individuals" -- the number of contacts per unit time is given by S * I. If you halve the population, then each has S * I / 2 -- effectively you are missing out on all the contacts between pop A individuals and pop B individuals. Because of this, the result is an exponential reduction in the growth. Another way to see this is if, instead, you broke the population up into n really small communities: then you'd be missing a massive number of potential contacts. I think you're right that this matters even more when you have more infectives, but even from the start, it represents a significant reduction.

Issue #70, opened 2022-01-07
Add a model allowing tracing and isolation


ben18785:
A key containment policy has been tracing and isolation of symptomatic individuals and their contacts. At the moment, we don't have this in there (although it could be argued that the NPIs model sorta does).

PR #73, opened 2022-01-08
Contact tracing and quarantining NPI model

ben18785:
Addresses #70 

No tests yet as awaiting feedback on model


R/SEIRD_CT.R
```diff
@@ -0,0 +1,305 @@
...
+#' @return an R0 value
+#' 
+#' @export
+setMethod("R0", "SEIRD_CT", function(model) {
+  beta_a = transmission_parameters(model)$beta_a
+  beta = transmission_parameters(model)$beta
+  kappa = transmission_parameters(model)$kappa
+  gamma = transmission_parameters(model)$gamma
+  mu = transmission_parameters(model)$mu
+  pi=transmission_parameters(model)$pi
```
solveigvandervegt: looks like pi is a protected name. Maybe consider renaming this?

R/SEIRD_CT.R
```diff
@@ -0,0 +1,305 @@
...
+  omega=transmission_parameters(model)$omega
+  eta_a=transmission_parameters(model)$eta_a
+  psi=transmission_parameters(model)$psi
+  phi=transmission_parameters(model)$phi
+  
+  infections_asymptomatics <- beta_a * eta_a / gamma
+  infections_presymptomatics <- beta / psi
+  infections_infectious <- (1 - phi) * beta / (gamma + mu)
+  
+  infections_asymptomatics + (1 - pi) * (1 - eta_a) * (
```
solveigvandervegt: I will trust you on this but I don't immediately see how this is the appropriate value for R0 for this model. Maybe I'll find it in the vignette but here in the documentation you just note R0 = beta/gamma+mu and that looks very different from what you are doing here.

vignettes/SEIRD_CT.Rmd
```diff
@@ -0,0 +1,362 @@
...
+# tmp = DiagrammeRsvg::export_svg(g)
+# tmp = charToRaw(tmp)
+# rsvg::rsvg_png(tmp, "images/seird_ct_full.png", width = 1000)
+knitr::include_graphics("images/seird_ct_full.png")
+```
+
+Effectively, this model has two parallel tracts, which are copies of the first model in this vignette: one representing unisolated individuals (i.e. those compartments without the "T") and those representing quarantining individuals (with a "T"). There are two flows into the quarantined tract: susceptibles can be isolated upon infection $S\rightarrow E^T$ if they are known contacts of an individual who goes on to be symptomatic: i.e. either $P$ or $I$; or infected individuals can be quarantined once they become symptomatic $P\rightarrow I^T$.
+
+## Basic reproduction number
+It can be shown that the basic reproduction number for this model is given by^[Following [@vandenDriessche17] (see `Vignette("SEIaImIsRD")` for more information) and assuming a state order: $(E, P, A, I, E^T, P^T, A^T, I^T)$, we obtain the matrices:
```
solveigvandervegt, line 109: I am not sure if the sequence by^[Following [@vandenDriessche17] does something in the knitted vignette that I just cannot see or if it is typo so highlighting anyway :)

vignettes/SEIRD_CT.Rmd
```diff
@@ -0,0 +1,362 @@
...
+
+The expression for $R_0$ shows that even if contact tracing and isolation of contacts is perfect ($\pi=1$), there remains transmission due solely to asymptomatic cases. A meta-analysis concluded that the overall proportion of cases due to asymptomatics is 17% [@byambasuren2020estimating], meaning that if the transmission rate of asymptomaics, $\beta_a$ is high enough (or, although this model does not include it, a longer duration of infection), it is possible to sustain an epidemic solely through asymptomatic transmission: if $\eta \beta_a/\gamma>1$. This analysis suggests that regular testing, regardless of whether an individual has symptoms, may be an effective way to target asymptomatic reservoirs of infection, if upon testing positive individuals and their close contacts isolate.
```
solveigvandervegt: Does this make sense? Cause if contact tracing is perfect, shouldn't you also identify all the contacts who become asymptomatic and be able to quarantine those? I guess that might depend on your protocol after you have identified a contact.

solveigvandervegt (commented):
Hey Ben, looks really good. A few small comments. 

This is more of a side note, but I do wonder if writing all of the functions (and functions inside functions) in the vignette is going to be super helpful for someone who is new to epidemiology and potentially new to R. Of course, there is great benefit to writing code this way but it takes a lot longer to understand the flow, especially for someone new to this. Since the vignette is something people actually read, as opposed to the classes that are more behind the scenes, I would tend to try and keep the code as simple and easy to understand as possible. No need to change anything, just something to think about.

R/SEIRD_CT.R
ben18785: Good idea: I've changed it to chi.

R/SEIRD_CT.R
ben18785: Good spot re: the docstrings R_0 value (which was wrong). The R0 value is ok I think -- it's all derived in the vignette footnotes and seems to make sense to me.

vignettes/SEIRD_CT.Rmd
ben18785: Yep, this is the footnote where I derive the R_0 value :)

vignettes/SEIRD_CT.Rmd
ben18785: Yes, contact tracing can find asymptomatic people (i.e. contacts of symptomatics). But contact tracing is unlikely to find contacts (symptomatic or otherwise) of asymptomatic people because so few asymptomatics are identified.

ben18785:
Yep, re:

> This is more of a side note, but I do wonder if writing all of the functions (and functions inside functions) in the vignette is going to be super helpful for someone who is new to epidemiology and potentially new to R. Of course, there is great benefit to writing code this way but it takes a lot longer to understand the flow, especially for someone new to this. Since the vignette is something people actually read, as opposed to the classes that are more behind the scenes, I would tend to try and keep the code as simple and easy to understand as possible. No need to change anything, just something to think about.

I completely agree. I think keeping long functions to a minimum is worthwhile. Here, there are a few, I admit, which are due mainly to a) finding a beta value that matches a given R0 value and b) running simulations across the combinations of variants and scenarios. It might be possible at some later date to make a function which automates finding a beta value to achieve a given R0, but let's leave that for now I reckon.

